PR #11, opened 2019-08-02
travis: use custom FQDN for GitHub pages

gwarf:
travis: use custom FQDN for GitHub pages: https://docs.egi.eu

Before one hour the Let's Encrypt certificate for HTTPS should be setup.


gwarf:
Please wait before merging, need to add some more static content (like a README and the github PR template that was dropped as we don't keep history not to keep old files).

PR #24, opened 2020-03-09
Notebooks documentation

enolfc:
# Summary 

----

**Related issue :**


enolfc:
Just taken from the existing docs in rtfd.io

content/users/Notebooks/_index.md
```diff
@@ -0,0 +1,57 @@
+---
+title: "Notebooks"
+linkTitle: "Notebooks"
+type: docs
+description: "Documentation related to EGI Notebooks"
+weight: 40
+---
+
```
gwarf:
```suggestion
```

enolfc:
> I've just made a lightening fast review, assuming it was already reviewed earlier. So only one small suggestion.

I did a bit of reorganisation, but content was indeed reviewed in previous PR for the other repo

> We may want to improve the automated checks to be sure we do all things in a coherent way (when there are various ways to write markdown or to use hugo short codes and so on).

I would expect mdl does force some coherent markdown... for the shortcodes, we should at least take a decision on how to include images and we should review relative links

gwarf:
> for the shortcodes, we should at least take a decision on how to include images and we should review relative links

I think we will need a clear contribution and style guide at some point in time...

PR #26, opened 2020-03-10
Accessing gocdb

gwarf:
Short documentation on accessing GOCDB, covering X509 and Check-in.
It's currently put inside the `Federated Operations` section, wasn't totally clear to me where to put it and I'm open to suggestions.


vardizzo-lab:
I see two possible options:

- In the Check-in user documentation, a paragraph is created that describes the procedure for requesting a role and then a sub-paragraph for the specific GocDB case. Then on FedCloud the link to that paragraph is used where necessary.

- The specific procedure for the GocDB role is described on the FedCloud documentation and the Check-in user documentation is left with the paragraph only describing how to request a role (other than GocDB, a specific one for a VO for example).

gwarf:
So I don't feel having GOCDB access under check-in would be much convenient/natural to users, they care about accessing the GOCDB not about using Check-in. I would prefer to have the GOCDB access doc page point to the check-in page when it make sense, as reference (if needed we can also have a reverse reference).
GOCDB access is not specific to FedCloud so I would not put it there.
The GOCDB access page would be linked from where it makes sense (FedCloud and other HTC services or maybe in the Getting Started section of the Users and Providers sections?).

gwarf:
Also in order to avoid having people to click many times to access the various parts of the doc maybe we could have a way to include the same doc at various places using something like https://gohugo.io/content-management/page-bundles/#headless-bundle?

content/operations/GOCDB/Accessing/_index.md
```diff
@@ -0,0 +1,54 @@
...
+[Users and
+roles](https://wiki.egi.eu/wiki/GOCDB/Input_System_User_Documentation#Users_and_roles)
+section.
+
+All roles applications need to be validated by parent roles or administrators.
+Once this is done, you can access/modify relevant information according to the
+role you have been granted. You can learn more on roles and user accounts by
+reading the #Users and roles section of this documentation.
+
+## Using an X509 digital certificate
```
enolfc: Should we prioritise Check-in here? I would put it first

content/operations/GOCDB/Accessing/_index.md
```diff
@@ -0,0 +1,54 @@
...
+Grid Certificate Profile document.
+
+## Using institutional account via EGI Check-in
+
+In order to be able to access the GOCDB with their institutional account, users
+need to:
+
+1. Have their Identity Provider (IdP) federated in eduGAIN ;
+1. Have created an [EGI Check-in account](https://wiki.egi.eu/wiki/AAI_usage_guide) ;
+1. Have requested joining the [GOCDB user
```
enolfc: Not really, if you have a fully-compliant with whatever R&S, SIRTIFI and whoknowswhatlelse, then it should work :)

content/operations/GOCDB/_index.md
```diff
@@ -0,0 +1,18 @@
...
+type: "docs"
+---
+
+{{% pageinfo %}}
+This documentation about the GOCDB is a work in progress,
+the main source of documentation about GOCDB is still https://wiki.egi.eu/wiki/GOCDB.
+{{% /pageinfo %}}
+
+Grid Configuration Database (GOCDB) contains general information about the
+sites participating to the production Grid. Accessed by all the project actors
```
enolfc:
```suggestion
sites participating to the production infrastructure. Accessed by all the project actors
```

enolfc: also project seems not the right word

content/operations/GOCDB/_index.md
```diff
@@ -0,0 +1,18 @@
...
+(end-users, site managers, NGI mangers, support teams, VO managers), by other
+tools and by third party middleware in order to get Grid topology. The portal
```
enolfc:
```suggestion
tools and by third party middleware in order to get the infrastructure topology. The portal
```

enolfc:
> So I don't feel having GOCDB access under check-in would be much convenient/natural to users, they care about accessing the GOCDB not about using Check-in. I would prefer to have the GOCDB access doc page point to the check-in page when it make sense, as reference (if needed we can also have a reverse reference).
> GOCDB access is not specific to FedCloud so I would not put it there.
> The GOCDB access page would be linked from where it makes sense (FedCloud and other HTC services or maybe in the Getting Started section of the Users and Providers sections?).

I agree with @gwarf  here, GOCDB deserves to be documented on its own. And probably is a very suited to the Getting Started section of the *providers*, users do not care so much about it.

content/operations/GOCDB/Accessing/_index.md
gwarf: I thought about ti too and was a bit unsure about this, as currently we are doing the opposite and the COU is a last resort.
Also ideally the X509 access should be done via Check-in IGTF IdP proxy, but while it works it will map people to a different user in GOCDB (if they already authentified via X509 to GOCDB)

content/operations/GOCDB/Accessing/_index.md
gwarf: Oh indeed, forgot about the fact that the COU is only for the people from "bad" IdPs, good catch, will update this.

content/operations/GOCDB/_index.md
gwarf: It's the introductotry blurb from the wiki.
We could just say `accessed by all the actors`?
Or a better/more clear proposition?

content/operations/GOCDB/Accessing/_index.md
gwarf: Implemented.

content/operations/GOCDB/Accessing/_index.md
gwarf: Updated.

gwarf:
> I agree with @gwarf here, GOCDB deserves to be documented on its own. And probably is a very suited to the Getting Started section of the _providers_, users do not care so much about it.

So currently the `Access` doc is still under a GOCDB directory, but is linked from the Getting Started. Does it look OK?

enolfc:
> So currently the `Access` doc is still under a GOCDB directory, but is linked from the Getting Started. Does it look OK?

Yes, for the time being it should be ok. For sure the Getting Started page needs to have some actual content, but that's another PR

gwarf:
@paolini78 I've added you as a reviewer, if you want to have a look and word before it gets merged

gwarf:
I will merge now, feel free to comment/update issue later.

PR #47, opened 2020-06-30
config: use lunar.js search instead of non working google search

### LINKED ISSUE #46: Fix or disable search field
The currently available search field is not working, example: https://docs.egi.eu/search/?q=training

Search should be fixed or disabled.

Docsy documentation: https://www.docsy.dev/docs/adding-content/navigation/

gwarf:
config: use lunar.js search instead of non working google search as documented on https://www.docsy.dev/docs/adding-content/navigation/#configure-local-search-with-lunr.
Fix #46

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -92,13 +92,13 @@ time_format_default = "January 2, 2006"
 rss_sections = ["blog"]
 
 # Google Custom Search Engine ID. Remove or comment out to disable search.
-gcs_engine_id = "011217106833237091527:la2vtv2emlw"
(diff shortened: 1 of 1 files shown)


gwarf:
> Thanks @gwarf! Seems good according to documentation

It works locally, but we will have to double check it works in prod without having to make any change to the build/deploy travis part.

gwarf:
It seems it works in prod on https://docs.egi.eu
Screenshot 2020-06-30 at 17 44 57

PR #70, opened 2020-10-02
Reorganisation

enolfc:
# Summary 

Move things around to match the discussion on how we should organise the docs

----

**Related issue :**


enolfc:
Missing the CODEOWNERS... getting there

gwarf (changes requested):
Great!
From a first check:
* remove the GitHub section, it should go under About as contributing, cf the doc
* split HTC over multiple pages

gwarf:
Maybe also better to remove the generic text from docsy like on http://localhost:1313/users/

enolfc:
> Maybe also better to remove the generic text from docsy like on http://localhost:1313/users/ or http://localhost:1313/users/getting-support/

but still leave the placeholder?

gwarf:
Either add some initial short text (better) or keep the placeholder (acceptable I would say)

andrea-manzi (approved):
All good from my side!

gwarf (dismissed):

CODEOWNERS
```diff
@@ -24,6 +24,9 @@ content/*/*/datahub/ @andrea-manzi
 # Data Transfer
 content/*/*/data-transfer/ @andrea-manzi
 
+# Data Transfer
```
andrea-manzi, line 27:
```suggestion
# Online Storage
```

gwarf (approved):
LGTM, thanks!

andrea-manzi (approved):
Good for me

PR #101, opened 2020-10-26
Update _index.md

glarocca:
Removed placeholder page

# Summary 

----

**Related issue :**


gwarf (changes requested):
Would it be possible to add some simple sentence so that at least the page is not empty?

enolfc:
Shall we close this one? as there is another just populating this page?

PR #123, opened 2020-11-03
getting started providers: route to various places

gwarf:
Some simple text to remove a placeholder.

Wondering if notebooks doc shouldn't go under internal services, is it really something to be deployed by service providers themselves?

### DIFF
--- a/content/en/providers/getting-started/_index.md
+++ b/content/en/providers/getting-started/_index.md
@@ -5,6 +5,17 @@ type: docs
 weight: 10
 ---
 
-{{% pageinfo %}}
-This is a placeholder page.
-{{% /pageinfo %}}
+The pages under this section depict how to join the EGI infrastructure as
+a service providers offering innovative services to the European Research Area.
+
+> Interested in integrating with Check-in? Head to the
+  [Check-in for service provicers](https://docs.egi.eu/providers/check-in/)!
+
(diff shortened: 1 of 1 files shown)


enolfc:
The notebooks docs is about the deployment of the instances, it's true that there is 1/2 providers, but where would you move it to?

content/en/providers/getting-started/_index.md
```diff
@@ -5,6 +5,17 @@ type: docs
 weight: 10
 ---
 
-{{% pageinfo %}}
-This is a placeholder page.
-{{% /pageinfo %}}
+The pages under this section depict how to join the EGI infrastructure as
+a service providers offering innovative services to the European Research Area.
+
+> Interested in integrating with Check-in? Head to the
```
enolfc, line 11: what about IdPs?

gwarf: the integrating was expected to cover both IdP and SP in fact. Maybe we could be more clear and point to the sub sections for both aspects.

gwarf: I've updated to cover this.

gwarf:
> The notebooks docs is about the deployment of the instances, it's true that there is 1/2 providers, but where would you move it to?

Let's keep it there for now.

enolfc (approved):
👍

PR #124, opened 2020-11-04
Add "User attributes" section

nikosev:
# Summary

Add the "User Atrtibute" section in "Service Providers" < "Check-in" < "Service Providers" guide.
The section provides a detailed description of the attributes that Check-in releases to the Service Providers.
Also this PR fixes the format of  current [Claims](https://docs.egi.eu/providers/check-in/sp/#claims) section.

---

**Related issue :** -


content/en/providers/check-in/sp/_index.md
```diff
@@ -291,99 +278,17 @@ value pairs for the Claims.
 The following scope values can be used to request Claims from the EGI
 Check-in UserInfo Endpoint:
 
+| Scope                          | Claims                                                                                                    |
+| ------------------------------ | --------------------------------------------------------------------------------------------------------- |
+| `openid`                       | `sub`                                                                                                     |
+| `profile`                      | - `name` <br> - `given_name` <br> - `family_name` <br> - `preferred_username`                             |
```
enolfc, line 284: maybe try with a <ul> instead of <br>? Not really sure what's best for maintenance of the docs / rendering

content/en/providers/check-in/sp/_index.md
```diff
@@ -769,6 +674,198 @@ Alternatively VO portals could implement such functionality themselves
...
+
+### 2. Display Name
+
+|          attribute name | Display Name                                      |
+| ----------------------: | :------------------------------------------------ |
+|         **description** | The user's full name, in a displayable form       |
+|   **SAML Attribute(s)** | `urn:oid:2.16.840.1.113730.3.1.241` (displayName) |
+|          **OIDC scope** | `profile`                                         |
+|       **OIDC claim(s)** | `name`                                            |
+| **OIDC claim location** | - Userinfo endpoint                               |
```
enolfc, line 707: why the `-` before Userinfo?

enolfc (commented):
Thanks @nikosev!

Just one thing on how to format the lists inside the tables as I'm not really sure on what's the best approach, @gwarf, any opinion here?

content/en/providers/check-in/sp/_index.md
nikosev: It suppose 'OIDC claim location' to have a list of locations.
Since for this case, there is only one I can remove it the `-`

content/en/providers/check-in/sp/_index.md
gwarf: > Just one thing on how to format the lists inside the tables as I'm not really sure on what's the best approach, @gwarf, any opinion here?

So by definition html code in markdown is not great (and making linters to complain)(yes we can disable some options, but it's suboptimal) and should be avoided as much as possible, but unfortunately markdown is also a bit limited for the advanced management of tables.

So in case we have not other option than using HTML, maybe using some tags containing more meanings would be more appropriate, like:
 ```html
<ul><li>Userinfo endpoint</li><li>introspection endpoint</li></ul>
```

Also when `<br>` are required it would be preferable to use the `<br />` syntax.

gwarf: Thanks @nikosev!

enolfc (approved):
LGTM

gwarf:
@nikosev so we are testing some [new linting setup](https://github.com/EGI-Foundation/documentation/pull/124/checks?check_run_id=1358153017
), but don't be worry we will accept this PR without having them all addressed, we will take care of this separately.

Another PR is on his way to adapt the link checking to skip the private AAI URLs and the example.com one.
Should we also ignore https://aai.egi.eu/LoA#Low?

nikosev:
> Should we also ignore https://aai.egi.eu/LoA#Low?

@gwarf yes please!

gwarf:
Oops: https://github.com/EGI-Foundation/documentation/pull/124/checks?check_run_id=1359645667#step:4:36
[
Screenshot 2020-11-05 at 18 27 22
](url)

gwarf:
So the refeds page is back, let's get this merged.

PR #125, opened 2020-11-04
github actions: add superlinter but ignoring errors

gwarf:
A PR replacing #123, let's start easy by not breaking on all errors, we will fine tune linter configuration and fix remaining errors section by section and enforce when it will be ready for checking all code.


gwarf:
Current output of the check: https://github.com/gwarf/documentation/runs/1353769093?check_suite_focus=true

.github/linters/.markdownlint.json
```diff
@@ -0,0 +1,7 @@
+{
+  "MD002": false,
+  "MD024": false,
```
enolfc, line 3: was this causing much trouble?

enolfc:
> Current output of the check: https://github.com/gwarf/documentation/runs/1353769093?check_suite_focus=true

🙈 
that's a long list of errors...

enolfc (approved):
I think this is a good approach, let's go incrementally

.github/linters/.markdownlint.json
gwarf: For [MD002](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md002---first-heading-should-be-a-top-level-heading) the fact is that AFAIU files should start with a second level header `##` as the first one would be more or less the page title and it's set by the title entry in the frontmatter.
In fact it seems I lost some changes somewhere, MD002 is now disabled by default as [MD041](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md041) is taking care of all this, so will drop it to make things easier/cleaner.

For the [MD024](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md024---multiple-headings-with-the-same-content) we had some pages including some repeated header title and that was legit.

PR #128, opened 2020-11-05
Note about working from CLI on a PR created online

gwarf:
Note about working from CLI on a PR created online.

cf. https://github.com/EGI-Foundation/documentation/pull/126#discussion_r517933195


enolfc (commented):
Could you separate stuff not really related to the change? (mlc config)

gwarf:
> Could you separate stuff not really related to the change? (mlc config)

Yeah it's dirty, but the fact is that it's required to have test pass, and moving those changes to another PR may not be worth the time cost.
Will refrain doing this next time, I'm hiding :)

PR #129, opened 2020-11-05
GitHub actions: cleaning and update of checks

gwarf:
GitHub actions: cleaning and update of checks:
* markdown-link-check: ignore some more links
* drop markdownlint usage
* fix building on GitHug actions, using npm with `package.json` and `package-lock.json`


enolfc (approved):
Looking good!

PR #136, opened 2020-11-06
merge workflows

enolfc:
# Summary

---

**Related issue :**


gwarf (commented):
So while removing duplications is definitely a good point I’m a bit vary about this, in that case we need to rename the workflow (As there is no deploy there for PR), maybe too smart/hackish, don't like much the condition…
I’m also still too much newbie with this to have strong opinions :)

enolfc:
Let's drop it, just wanted to avoid too much duplication

PR #141, opened 2020-11-17
Reformat Check-in files

nikosev:
# Summary

Reformat Check-in files using VScode's Prettier and markdownlint extensions


gwarf (changes requested):
Thanks @nikosev, so quickly looking it seems it's still missing mainly addressing the long lines due to the big tables.
For this we advise two things:
- moving to something else than a table when a table doesn't really makes sense
- when a table is indeed the proper way to represent this data, it can be encapsulated with some tags to disable markdown-lint checks:

```markdown
<!-- markdownlint-disable line-length -->
| Action           | rOCCI                                                                                                                | OpenStack                                                                                |
| ---------------- | -------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------- |
| List images      | `occi -a list -r os_tpl`                                                                                             | `openstack image list`                                                                   |
<!-- markdownlint-enable line-length -->
```

content/en/providers/check-in/sp/_index.md
```diff
@@ -180,103 +163,92 @@ Service Providers is included in the [User Attribute](#user-attributes) section.
...
 
-- [Shibboleth SP
-  Configuration](https://wiki.shibboleth.net/confluence/display/SHIB2/NativeSPConfiguration)
-- [SimpleSAMLphp Service Provider
-  QuickStart](https://simplesamlphp.org/docs/stable/simplesamlphp-sp)
-- [Simple SAML 2.0 service provider with mod_auth_mellon Apache
-   module](https://github.com/UNINETT/mod_auth_mellon)
+- [Shibboleth SP Configuration](https://wiki.shibboleth.net/confluence/display/SHIB2/NativeSPConfiguration)
+- [SimpleSAMLphp Service Provider QuickStart](https://simplesamlphp.org/docs/stable/simplesamlphp-sp)
+- [Simple SAML 2.0 service provider with mod_auth_mellon Apache module](https://github.com/UNINETT/mod_auth_mellon)
```
nikosev, line 181: @gwarf should I disable the line length check for the links too?
for example in these lines

content/en/users/check-in/voms/_index.md
```diff
@@ -252,34 +259,33 @@ existing VOs
...
 
-``` {.console}
+```{.console}
 voms-proxy-init --voms <name of the vo> --rfc
 ```
 
 See for example, using `fedcloud.egi.eu` VO:
 
-``` {.console}
+```{.console}
```
gwarf, line 288: Please use proper markdown fenced code markers (everywhere in the doc), like;
```
```sh
```

(sh is usually preferred over console as it highlights more stuff).

content/en/providers/check-in/sp/_index.md
gwarf: so I think/hope that for links it shouldn't be a problem as we are not expected to shorten them/split them over lines, but let me double check.

gwarf:
You may also have to also disable `no-inline-html` for the HTML code that was added into tables.

content/en/providers/check-in/sp/_index.md
gwarf: We will see how it works once the tables will have been fixed, I expect it's the main problem.

nikosev:
Thank you @gwarf for your comments. I have resolved the errors!

content/en/providers/check-in/sp/_index.md
gwarf: Great, so all seems fine with long URL.

gwarf (approved):

gwarf:
Perfect, thanks!
Let's get this merged.

PR #148, opened 2020-11-19
Only use the main support section

gwarf:
Only use the main support section, suggesting to use Helpdesk for sevice-specific support and having support email being available.


enolfc (approved):
LGTM

PR #153, opened 2020-11-23
Format Check-in files vol.2

nikosev:
# Summary

This PR contains the following changes:

- Add new line before and after "markdownlint" tag
- Remove "mailto" link
- Remove forgotten bracket in entitlement syntax
- Remove unnecessary indent
- Continue numbering in numbered lists
- Use production endpoints in examples

### DIFF
--- a/content/en/providers/check-in/idp/_index.md
+++ b/content/en/providers/check-in/idp/_index.md
@@ -38,10 +38,12 @@ The most important URLs for each environment are listed in the table below but
 more information can be found in the protocol-specific sections that follow.
 
 <!-- markdownlint-disable line-length -->
+
 | Protocol       | Development environment                                            | Demo environment                                                    | Production environment                                         |
 | -------------- | ------------------------------------------------------------------ | ------------------------------------------------------------------- | -------------------------------------------------------------- |
 | SAML           | <https://aai-dev.egi.eu/proxy/module.php/saml/sp/metadata.php/sso> | <https://aai-demo.egi.eu/proxy/module.php/saml/sp/metadata.php/sso> | <https://aai.egi.eu/proxy/module.php/saml/sp/metadata.php/sso> |
 | OpenID Connect | See [client registration](#client-registration)                    | See [client registration](#client-registration)                     | See [client registration](#client-registration)                |
+
 <!-- markdownlint-enable line-length -->
 
 ## General requirements for integrating identity providers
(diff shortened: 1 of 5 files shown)


content/en/providers/check-in/idp/_index.md
```diff
@@ -28,7 +28,7 @@ The integration follows a two-step process:
    affecting the production environment of EGI. Note that the development
    instance is not connected to the production service and no information is
    shared between the two systems.
-1. Register your Identity Provider with the production instance of EGI Check-in
+2. Register your Identity Provider with the production instance of EGI Check-in
```
gwarf, line 31: Usually it's better to rely on auto-numbering, is there any reason for introducing this manual numbering here?

gwarf (commented):
Thanks!
good to see some spacing coming to separate the markdownlint tags.
But I'm a bit surprised by the manual numbering, is there any reason for this?
Usually it's cleaner/easier to stick with the automatic numbering (ie. using 1. everywhere)

nikosev:
@gwarf You are right, I'm reverting the change about numbering

content/en/users/check-in/vos/_index.md
```diff
@@ -91,9 +91,8 @@ The VO enrollment URL can be found through the EGI Check-in Registry:
 ### Managing VO groups
 
 VO groups can only be created by Check-in platform administrators. Please
-contact [Checkin Support](mailto:checkin-support@mailman.egi.eu) indicating the
-following information for every (sub)group that you need to add/remove to/from
-your VO:
+contact `checkin-support` `<AT>` `mailman.egi.eu` indicating the following
```
gwarf, line 94: So here and there we have been using directly the email with a mail-to, usually our addresses are already quite easily findable on the net, not sure it's totally required to do a trick like this to try to avoid being spammed.

gwarf: Or @nikosev do you feel you have been receiving too many spam that wasn't blocked by mailman and so on?

nikosev: @gwarf We get on average 5 spam mails per month on this email address, which create automatically GGUS tickets. But if you think it is pointless to try to protect it, then I can revert this change

gwarf: OK, fine with me to keep it like this, just not sure how effective this protection will be.

gwarf (approved):
LGTM

PR #170, opened 2020-12-23
Updated the FGSG url

glarocca:
# Summary

---

**Related issue :**


enolfc (commented):
@glarocca I'd prefer not to change the lists in this change. The rendering will be the same whether we use `-` or `*`. I have updated #167 to re-apply your change on the reformatted files. If you are ok with that one, I prefer to merge that and close this

PR #172, opened 2020-12-23
Fix linter issues

enolfc:
# Summary

---

**Related issue :**

### DIFF
--- a/.github/linters/mlc_config.json
+++ b/.github/linters/mlc_config.json
@@ -19,7 +19,7 @@
       "pattern": "^https://argo-mon.*.egi.eu/nagios"
     },
     {
-      "pattern": "^http://training.notebooks.egi.eu*"
+      "pattern": "^https://training.notebooks.egi.eu*"
     }
   ]
 }
--- a/.github/linters/mlc_config.json
+++ b/.github/linters/mlc_config.json
@@ -17,6 +17,9 @@
     },
     {
       "pattern": "^https://argo-mon.*.egi.eu/nagios"
+    },
+    {
+      "pattern": "^http://training.notebooks.egi.eu*"
     }
   ]
 }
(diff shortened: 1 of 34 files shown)


enolfc:
Still missing wrong links in one file. See https://github.com/EGI-Foundation/documentation/pull/172/checks?check_run_id=1604825797#step:4:35

no idea on how to make this work 🤷

gwarf:
> Still missing wrong links in one file. See https://github.com/EGI-Foundation/documentation/pull/172/checks?check_run_id=1604825797#step:4:35
> 
> no idea on how to make this work 🤷

So I feel like the markdown link checker may be interpreting the link incorrectly, as it's not proper markdown but an hugo shortcode.
Mabye we don't have much other solutions than adding an exception for this content. Or play with the [replacement option](https://github.com/tcort/markdown-link-check#config-file-format) to drop the quote, but not sure it's a good/clean idea.

enolfc:
> > Still missing wrong links in one file. See https://github.com/EGI-Foundation/documentation/pull/172/checks?check_run_id=1604825797#step:4:35
> > no idea on how to make this work 🤷
> 
> So I feel like the markdown link checker may be interpreting the link incorrectly, as it's not proper markdown but an hugo shortcode.
> Mabye we don't have much other solutions than adding an exception for this content. Or play with the [replacement option](https://github.com/tcort/markdown-link-check#config-file-format) to drop the quote, but not sure it's a good/clean idea.

What I don't understand is why the egi.eu and docs.egi.eu links fail but not the twitter one. I can add an exception for exactly those two links with `&quote;`

gwarf (approved):
LGTM, thanks!

andrea-manzi (approved):
thanks  a lot!

PR #194, opened 2021-03-22
Update references to EGI-Foundation, and use EGI-Federation

gwarf:
Update references to EGI-Foundation, and use EGI-Federation as the GitHub organisation has been renamed.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -60,6 +60,6 @@ jobs:
         with:
           # Use a personal token configured as a repository-specific secret
           personal_token: ${{ steps.generate-token.outputs.token }}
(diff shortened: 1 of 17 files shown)


gwarf:
From what I can see the check links errors are false negative, so @enolfc we can probably ignore them.

enolfc:
> From what I can see the check links errors are false negative, so @enolfc we can probably ignore them.

mmm, nope the notebooks ones need to be updated
and the other one is a bit puzzling...

gwarf:
> > From what I can see the check links errors are false negative, so @enolfc we can probably ignore them.
> 
> mmm, nope the notebooks ones need to be updated
> and the other one is a bit puzzling...

So it's not failing at the same place, before it was lots of false negatives for the appdb wiki. Will update the one that are indeed broken.

PR #197, opened 2021-03-23
Update Check-in FAQ section

ioigoume:
# Summary

This Pull Request extends `FAQ` guide in the User guide of Check-in


gwarf:
Hi @ioigoume
As it seems it's a false negative, could you add `https://aai.egi.eu/proxy/saml2/idp/SingleLogoutService.php?ReturnTo=https%3A%2F%2Faai.egi.eu%2Fregistry%2Fauth%2Flogin` to the whitelist list in https://github.com/EGI-Federation/documentation/blob/master/.github/linters/mlc_config.json ?

The Linting errore seems to be due to super-linter, let's see if it gets fixed, otherwise we will do a manual check and for the merge.

ioigoume:
Hi @gwarf ,
i made the fix and now two out of three pass.

Could you please check if we need to make a manual check for the last one.

-Ioannis

gwarf (approved):
LGTM, thanks!
Will bypass super-linter that is failing for an unknown/unclear reason.

PR #202, opened 2021-03-25
Full git history needed for proper list of changed files

gwarf:
Full git history is needed to get a proper list of changed files within `super-linter`.
As documented on https://github.com/github/super-linter#example-connecting-github-action-workflow
Should address issues with super-linter not able to find all the changes.


enolfc (approved):
LGTM

PR #212, opened 2021-05-12
Rename main to master

gwarf:
Rename main to master, update references in workflows and conf.


enolfc (approved):
:+1:  Looks good, I assume all instances of master are here

PR #219, opened 2021-05-26
lint: upgrade to slim-v4

gwarf:
Slim version do not include rust, dotenv, armttk and pwsh: https://github.com/github/super-linter#slim-image

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -33,7 +33,7 @@ jobs:
 
       # Runs the Super-Linter action
       - name: Run Super-Linter on all codebase
-        uses: github/super-linter@v3
+        uses: github/super-linter@slim-v4
         env:
           DEFAULT_BRANCH: main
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -33,7 +33,7 @@ jobs:
 
       # Runs the Super-Linter action
       - name: Run Super-Linter on all codebase
-        uses: github/super-linter@v3
+        uses: github/super-linter@slim-v4
```
enolfc, line 36:
```suggestion
        uses: ghcr.io://github/super-linter@slim-v4
```
That's what's written in the documentation ;)

enolfc (changes requested):

.github/workflows/lint.yml
gwarf: Not sure why their doc is using this only for the slim, it's a bit confusing. Thanks for spotting this!

PR #230, opened 2021-06-11
Updated users/cloud compute/openstack section

thebe14:
# Summary

---

**Related issue :**


content/en/users/cloud-compute/openstack/_index.md
```diff
@@ -29,14 +29,39 @@ of this guide will focus on CLI/API access.
...
 pip install requests
 pip install openstackclient
 ```
 
-- `jq` to easily manage JSON output
+To be able to use sed-like commands to map, filter, slice, and transform
+structured data (JSON), install `jq`:
+
+```shell
+pip install jq
```
enolfc, line 60: Is this a python tool? you need to install the binary before (e.g. apt-get install jq)

content/en/users/cloud-compute/openstack/_index.md
```diff
@@ -29,14 +29,39 @@ of this guide will focus on CLI/API access.
...
+> - Windows 10 SDK (`<latest`>)
+>
+> In case you prefer to use non-Microsoft alternatives for building non-pure
+> packages, please see [here](https://wiki.python.org/moin/WindowsCompilers).
+
+Installation of the OpenStack client can be done using:
 
 ```shell
+pip install --upgrade setuptools
 pip install requests
```
enolfc, line 52: Probably this also should be removed, it comes as a dependency

content/en/users/cloud-compute/openstack/_index.md
```diff
@@ -29,14 +29,39 @@ of this guide will focus on CLI/API access.
...
+> - Testing tools core features - Build Tools
+> - Windows 10 SDK (`<latest`>)
+>
+> In case you prefer to use non-Microsoft alternatives for building non-pure
+> packages, please see [here](https://wiki.python.org/moin/WindowsCompilers).
+
+Installation of the OpenStack client can be done using:
 
 ```shell
+pip install --upgrade setuptools
```
enolfc, line 51: is this windows specific?

enolfc (commented):
Thanks @thebe14! Some comments go inline

content/en/users/cloud-compute/openstack/_index.md
```diff
@@ -29,14 +29,39 @@ of this guide will focus on CLI/API access.
 
 ## Installation
 
-TBC
+The OpenStack client is a command-line client for OpenStack that brings the
+command set for Compute, Identity, Image, Object Storage and Block Storage
+APIs together in a single shell with a uniform command structure.
+
+> If installing on Windows, as there are non-pure Python packages needed for
```
enolfc, line 36: how does this look? is this a quote?

content/en/users/cloud-compute/openstack/_index.md
thebe14: yes, that is a python tool. however, it is not a pure Python package, so if you try to install it using pip it will fail to compile the non-Python parts. at least on my Windows machine. thus the authors also offer it in prebuilt form for installation on windows.

content/en/users/cloud-compute/openstack/_index.md
thebe14: apparently, a lot of people had problems installing the openstackclient package, because of two things: they did not have the  ms build tools, and then they had an older python installation that was not able to properly build the non-pure packages required for openstack. this ensures that the setuptools are up to date. sure, if you start from a fresh Python installation, you don't need this.

content/en/users/cloud-compute/openstack/_index.md
thebe14: it is a blockquote. check it out at https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#blockquotes. however, I was worried that this will be broken by the linter requirement on this repo to cap lines at max 80 chars. apparently, it merges successive such quotes into a single one, when rendered.

content/en/users/cloud-compute/openstack/_index.md
thebe14: yes, I put Windows specific steps in blockquotes

content/en/users/cloud-compute/openstack/_index.md
thebe14: and the reference to it is already on the page. I just explained what it actually is and who might need it.

gwarf: So AFAICS it's python binding for jq, not the program itself.
More generally I'm not sure that everyone will want to install software via pip if there is another easier/cleaner/prefered option, like on linux people will probably prefer to use the OS package management (to keep it cleanly installed/updated/removed), and on Mac I've always been using homebrew for this.

I would prefer mentioning https://stedolan.github.io/jq/download/.

content/en/users/cloud-compute/openstack/_index.md
gwarf: Yes, it's a nice/handy quote, that's fine to use this I would say. (Often using it in Slack)

content/en/users/cloud-compute/openstack/_index.md
gwarf: I would remove the steps about setuptools upgrade, it doesn't seems like something we want to cover in this docs, I would stick the minimal and even more as here those are OS-agnostic steps.

content/en/users/cloud-compute/openstack/_index.md
gwarf: I would also remove the explicit request install, but if you feel you could mention something about problems/solutions in the quote for windows. 
Or maybe adding a subsection covering some specificities when installing on windows (what would be nice would be some tabs allowing to have // doc for installing on Windows, Linux or Mac that the user could select, but I don't think we currently have an easy/simple way to do this).

I think all this probably haven't been tested much on Windows, so it's certainly helpful to have this be done.

gwarf (changes requested):
Thanks, I've added some comments, similar to the ones @enolfc made, please have a look!

content/en/users/cloud-compute/openstack/_index.md
thebe14: done

content/en/users/cloud-compute/openstack/_index.md
thebe14: done

content/en/users/cloud-compute/openstack/_index.md
gwarf: I am still unsure that `pip install jq` should be mentioned/recommended.
It's python bindings not the jq binary, it's meant to interact with jq using python (see https://github.com/mwilliamson/jq.py), not much to do with using the jq CLI.
I would just fully remove the pip install jq stuff and only keep the link to the jq binary homepage.

thebe14: it is currently mentioned in the docs. actually, it's the only way mentioned, without any explanation why one would need it. but ok, removing the pip installation.

gwarf: It's not really what I see on https://docs.egi.eu/users/cloud-compute/openstack/#installation.
I see that it's mentioned that it can be used to manage json output, and not mention of installing it via pip, it's one of your commit that introduced this.
Screenshot 2021-06-30 at 09 14 53

The fact is that `jq` is never used after in the examples, so it feels a bit weird to see it mentioned without any example how it could be used/where it's useful.
Either we should provide some real example on how to use it to parse/format/filter some json output or maybe just removing it from there.

gwarf: See https://github.com/EGI-Federation/documentation/search?q=jq

gwarf: (Now I think that removing it from there is probably the cleanest thing to do)

thebe14: removed

gwarf (approved):
LGTM, thanks for providing some content for this!

enolfc (approved):
👍 Thanks for this!

PR #245, opened 2021-07-12
Updates modules after having checked out a PR

gwarf:
Update examples to updates modules after having checked out a PR


enolfc (approved):
👍

PR #246, opened 2021-07-12
Add links to fedcloudclient

sebastian-luna-valero:
# Summary

I added links to https://fedcloudclient.fedcloud.eu/ across all pages that refer to the tool.

I also added a minor subsection to emphasize that running the `fedcloud` command is particularly easy from EGI notebooks due to the existing integration with access tokens from EGI Check-in. Please note that the current default environment in EGI notebooks does not come with the `fedcloud` client installed, would this be a good idea?

At the moment here is what I do to have a permanently installed version of `fedcloud` in my EGI notebooks environment:
```bash
# First time round:
curl -O https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
source conda_deactivate.sh
bash Miniconda3-latest-Linux-x86_64.sh -b -p conda-install
source conda-install/etc/profile.d/conda.sh
conda install mamba -c conda-forge --yes
mamba create -n fedcloudclient python --yes
conda activate fedcloudclient
pip install fedcloudclient
export OIDC_ACCESS_TOKEN=`cat /var/run/secrets/egi.eu/access_token`
fedcloud token check

# After first setup, activation is as follows:
source conda_deactivate.sh
source conda-install/etc/profile.d/conda.sh
conda activate fedcloudclient
export OIDC_ACCESS_TOKEN=`cat /var/run/secrets/egi.eu/access_token`
fedcloud token check
```
To avoid conflicts with the existing `conda` I source `conda_deactivate.sh` beforehand:
```bash
#!/usr/bin/env bash

  EXISTS_CONDA=$(which conda) || $(echo "")

  if [[ "${EXISTS_CONDA}" ]] ; then
    CONDA_PATH=$(dirname $(dirname ${EXISTS_CONDA}))
    PATH_LIST=$(echo ${PATH} | tr ':' ' ')
    NEW_PATH=/usr/local/bin

    # loop to get rid of conda paths in PATH
    for e in $(echo ${PATH_LIST});
    do
      if [[ "${e}" != ${CONDA_PATH}* ]] && [[ "${e}" != "/usr/local/bin" ]] ; then
        NEW_PATH=${NEW_PATH}:${e}
      fi
    done

    # reconfig PATH
    PATH=${NEW_PATH}

    # loop to unset CONDA environment variables
    for e in $(env | grep CONDA | sed 's/=.*//g');
    do
      unset ${e}
    done

  fi
```

---

**Related issue :**


andrea-manzi:
Hi Sebastian
thanks!
have you tried to install the fedcloud client on the  existing conda  environment?  cause i think it's a good idea to  preinstall it on our notebooks images, but i would avoid in case there are conflicts with the existing env
cheers
Andrea

andrea-manzi:
ok i just did a quick test and it seems fine !

andrea-manzi:
ok i see that you have already added the possibility to use the client inside notebooks in the doc, so we should make this change before upgrading the documentation

sebastian-luna-valero:
Thanks, Andrea!

Indeed, the number of dependencies for `fedcloudclient` is high so I am not sure what's best:
* Provide instructions in the docs (as above) to install it manually
* Pre-install `fedcloudclient` in the default environment
* Provide a separate environment with `fedcloudclient` pre-installed

Best regards,
Sebastian

andrea-manzi:
i think that we should have it preinstalled cause ATM when you install a new software either on the default env or on a separated one this is not persisted (only  new env under the user home will be persisted).

let me create a PR and then we can discuss also with @enolfc about it

andrea-manzi:
so.. i have tried to add the fedcloudclient to the base notebook image and build it locally..and now the image size is 21 GB....i think it's better to add instructions but avoid to install it for now ( we should slim a bit that image), what do you think?

sebastian-luna-valero:
Thanks for double checking.

Let's wait for Enol's opinion but I am happy to just add the instructions for now.

enolfc (approved):
👍  I think it would be nice to have fedcloudclient installed at the notebooks by default, but need to find the right way to do so.

gwarf:
@sebastian-luna-valero please rebase and fix the conflict with the latest changes from main.

sebastian-luna-valero:
Done, thanks!

PR #259, opened 2021-07-23
Expiration policy guidelines for Check-in

ioigoume:
# Summary
Expiration policy guidelines for AAI Check-in Registry VO membership lifecycle

### DIFF
--- a/content/en/users/check-in/vos/_index.md
+++ b/content/en/users/check-in/vos/_index.md
@@ -231,6 +231,19 @@ A COU admins group owner can manage the admins group member as follows:
 
    ![Groups actions](./check-in-groups-actions.png)
 
+### Expiration Policy
+
+VO membership expires within a given time, typically a year after the VO member
+joins the VO. VO members receive a notification from [EGI Check-in Notifications](mailto:noreply@faai.grnet.gr)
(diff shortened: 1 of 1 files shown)


vardizzo-lab (changes requested):
The subject of the notification still seems to be defined ("TODO").
For the cycle of notifications, I would propose a different scheme than the one previously agreed, and which foresees the sending of notifications starting from one month before the expiry of the membership and on a weekly basis (no longer daily as previously). Only the last week before the deadline, I would activate on a daily basis. 
In order not to make the "user who does not want to renew the membership" passive upon receiving the notification emails, it would be possible to provide an option that allows him to express the fact that he does not want to renew the membership and that from that moment on he does not will receive more notifications?

ioigoume:
> The subject of the notification still seems to be defined ("TODO").
Fixed

vardizzo-lab (approved):
So we choose the weekly reminder solution with a message that contains, however, the number of days that remain to validate the membership. This notification mechanism also allows us not to disturb "too much" a user who has no intention of continuing to be a member of a VO and thus change his status (since dis-enrollment is not possible). 
Before this change, we communicate to the end-user the final day of its membership. We will alter this with the number of days remaining to expiration.

PR #266, opened 2021-07-30
Link to marketplace updated to EOSC

crsndr:
# Summary

---

**Related issue :**

### DIFF
--- a/content/en/users/cloud-compute/faq/_index.md
+++ b/content/en/users/cloud-compute/faq/_index.md
@@ -12,7 +12,7 @@ description: >
 
 There is a VO available for 6 months piloting activities that any researcher in
 Europe can join. Just place an order into the
-[EGI Marketplace](https://marketplace.egi.eu/31-cloud-compute).
+[EGI Marketplace](https://marketplace.eosc-portal.eu/services/egi-cloud-compute).
 
 ### How can I get an OAuth2.0 token?
 


gwarf (commented):
Thanks @crsndr, I'm a bit surprised, was it discussed/agreed that we should point to EOSC Marketplace instead of EGI Marketplace (that we control) for offering/presenting our services?

gwarf (changes requested):
I would first like to get a confirmation that it was agreed/decided to switch links to point to the EOSC MP and not to the EGI one.

crsndr:
@gwarf yes we need to check as after talking with @sebastian-luna-valero I understood that this is still under discussion. Who do you think is the best person to ask?

gwarf:
> @gwarf yes we need to check as after talking with @sebastian-luna-valero I understood that this is still under discussion. Who do you think is the best person to ask?

Would suggest to discuss this with Diego.

crsndr:
Ok. I think he's out of office for this week so if you could keep this on hold for now.

sebastian-luna-valero:
I discussed this with @glarocca a couple of weeks ago.

There are ongoing discussions to upgrade the EGI Marketplace so I think that we need to keep the existing link as is.

gwarf:
> I discussed this with @glarocca a couple of weeks ago.
> 
> There are ongoing discussions to upgrade the EGI Marketplace so I think that we need to keep the existing link as is.

Indeed. So let's close this PR and open a new one later if needed (and updating all references to the MP to point to the EOSC one if it was decided).

PR #270, opened 2021-08-06
Revised online storage user documentation

thebe14:
# Summary

Moved the block storage documentation under section Online Storage
Added detailed description of all block storage volume commands
Revised the description of usage of block storage volumes from inside of a VM
Revised the grid storage documentation
Revised the object storage documentation

---

TODOs Move (some of) the "Access..." chapters from each storage type to the relevant page under Tutorials


content/en/users/online-storage/_index.md
```diff
@@ -4,34 +4,37 @@ linkTitle: "Online Storage"
...
-comprises different categories of storage services depending on the technology
-and usage that is foreseen. The 3 following service offerings are available:
+## Overview
 
-- [Grid Storage](grid-storage/)
-- [Object Storage](object-storage/)
-- [Block Storage](../cloud-compute/storage/#block-storage), which is described
-  under the Cloud Compute section
+EGI Online Storage includes a wide category of services that allow
+users to **store, share and access data in the cloud**, using the EGI Cloud
```
andrea-manzi, line 13: i would not use the phrase "store, share and access data in the cloud," as the Grid Storage is mostly for HTC

andrea-manzi: the same for "using the EGI Cloud infrastructure" better to remove Cloud

content/en/users/online-storage/block-storage/_index.md
```diff
@@ -0,0 +1,662 @@
...
+### Attach volume to VM
+
+Mapping block devices to VMs is described in detail in the
+[OpenStack documentation](https://docs.openstack.org/nova/wallaby/user/block-device-mapping.html).
+
+To
+[attach a volume to a VM](https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/server.html#server-add-volume)
+use the FedCloud command below:
+
+{{% alert title="Note" color="info" %}} To be able to attach a volume to a VM,
```
andrea-manzi, line 225: there is no need to stop a VM to attach a volume, this can be done on the fly

content/en/users/online-storage/block-storage/_index.md
```diff
@@ -0,0 +1,662 @@
...
+argument to specify the device name in the VM should not be used. It does
+not work properly, and will be removed in the near future.
+{{% /alert %}}
+
+```shell
+$ fedcloud openstack server add volume my-server my-volume
+Site: IN2P3-IRES, VO: vo.access.egi.eu, command: server add volume my-server my-volume
+```
+
+You can check that the volume got attached o the VM (and with what device name)
```
andrea-manzi, line 240:
```suggestion
You can check that the volume got attached to the VM (and with what device name)
```

content/en/users/online-storage/block-storage/_index.md
```diff
@@ -0,0 +1,662 @@
...
+When the volume status is `in-use` the volume is attached to a VM, and it can
+be [used from the VM](#access-from-your-vms).
+
+### Detach volume from VM
+
+To
+[detach a volume from a VM](https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/server.html#server-remove-volume)
+use the FedCloud command below:
+
+{{% alert title="Note" color="info" %}} To be able to detach a volume from a
```
andrea-manzi, line 264: same as above

andrea-manzi (changes requested):
thanks a lot for the improvements to the doc! please check the few comments i made, then we are good to go

content/en/users/online-storage/_index.md
thebe14: But grid storage is still accessed over the internet, right? The storage is offered by (mostly OpenStack) IaaS providers, right? So it is in the cloud.

content/en/users/online-storage/block-storage/_index.md
thebe14: The commands I tried were failing if the VM was running.

content/en/users/online-storage/block-storage/_index.md
thebe14: Fixed

content/en/users/online-storage/block-storage/_index.md
thebe14: Fixed

content/en/users/online-storage/_index.md
andrea-manzi: The providers of grid storage + HTC could be also part of the Fedcloud and offer OpenStack as well, but it does not mean that the storage is provisioned via VMs ( most of the time is via bare metal nodes).

thebe14: I understand. But I still think of that as a cloud service. It is the same with commercial cloud service providers. Everything that is not on-prem is a cloud service. The expression "allow users to store, share and access data in the cloud" is generic, and covers all storage solutions EGI offers, then for each of the services offered you have a more specific summary under "What is it?" of the storage service's page. . But if you still think this is inaccurate, I can just drop the term "cloud".

gwarf: I think that in our community grid storage is definitely expected to be different from the cloud stuff. I would also be in favour of being more generic here.

thebe14: So more generic means without the term "cloud"?

thebe14: Fixed

andrea-manzi (approved):
great thanks a lot!

content/en/users/online-storage/object-storage/_index.md
```diff
@@ -140,34 +244,29 @@ $ fedcloud openstack --site IFCA-LCG2 --vo vo.access.egi.eu ec2 credentials crea
 +------------+------------------------------------------------------------------------------------------------------------------------------------------+
 ```
 
-<!-- markdownlint-enable line-length -->
```
gwarf, line 143: The line length exception is mainly just for tables, not for code.
Please keep it here, and for the shell excerpt use line returns + backslash to split over multiple lines.

content/en/users/online-storage/object-storage/_index.md
```diff
@@ -2,127 +2,231 @@
...
 | Name      |
 +-----------+
 | file1.txt |
 +-----------+
 ```
 
-Download an object:
+#### Download (the content of) an object
+
+<!-- markdownlint-disable commands-show-output -->
```
gwarf, line 180: Sorry for being a PITA, but it's the first time we encounter the need to disable `commands-show-output`, I'm also a bit puzzled with this, I feel adding the `$` makes things more noisy and less convenient for copy and pasting, so I would suggest to remove them as most of the time shell excerpt do not include them.
We can also document this in the style guide so that we have a clear rule for this.

gwarf (changes requested):

content/en/users/online-storage/object-storage/_index.md
thebe14: I understand. but I still think consistency is **much more important** for the reader. seeing the $ (or whatever we chose as the prompt) in that code block makes it clear that it is a command to run in a terminal. there are other such blocks in the docs that are not shell commands (and I intend to corrected them all). this so-called "rule" in my opinion is an arbitrary opinion of the linter authors, and is wrong/stupid (there are others as well).

if you want an example backing my opinion, try reading the docs of the commercial cloud providers. an example is the AWS docs, e.g. here https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-lamp-amazon-linux-2.html

however, if I get outvoted on this, I will do as you say.

content/en/users/online-storage/object-storage/_index.md
thebe14: I disagree. shell commands are not code. if I break them using backslashes just for readability, that means those lines will stay broken, irrespective of how the browser is resized.

thebe14: what I could do is wrap each shell example in individual disable/enable html comments (to keep the rule active for regular text. 

although, in my opinion there are some linter rules that are enforced now but shouldn't (the line length and no space at line end are two of them). the contributions should be made as easy as possible, and only enforce those rules that are important (e.g. no broken links - but even for this we don't have the right tools: the link checker is dumb and does not understand how Hugo renders content, which forces us to use sections instead of pages for everything, and even fails to validate perfectly valid links in at least one case, where I had to disable link checking).

gwarf: Long lines in shell excerpt are not really a good/usual practice and not everyone is using a big screen, "broken" lines is the standard way to render them.

People are also editing the source of files in terminals or small editors and unlimited lines are definitely not handy/clean.

We are also relying on prettier so that we are using quite standard and known style rule and don't have to discuss every rules, and people can usually just use prettier.

If you have another link checker to propose that work best we will be happy to switch to it, but until we have a better alternative we will stay on this one as it already helps more than having nothing at all.

thebe14: > Long lines in shell excerpt are not really a good/usual practice and not everyone is using a big screen, "broken" lines is the standard way to render them.

I agree. that is why the commands itself are as short as possible. only the command output in my examples is wide. breaking that would be wrong and misleading.

thebe14: > If you have another link checker to propose

unfortunately, I don't have a better one. but link checking is very important, and it makes sense to struggle with the tool we have. some of the other rules, however, are a different story.

content/en/users/online-storage/object-storage/_index.md
gwarf: I agree that we need to have a coherent practice for this, until now everyone was on his own, that's why I said we need a clear rule.
The nice thing with the AWS doc is that the copy button is not copying the prompt.
I'm fine with enforcing having a prompt indicator if others also feel it's working better for the readers.
In order to move forward with this PR I would suggest you keep them and later we confirm this with others, update the style guide and plan to review all the pages for this.

content/en/users/online-storage/object-storage/_index.md
thebe14: I gave up on prettier, it messes up the formatting in some cases. but more importantly, it makes reading the docs in a text editor a pain (after it applies rigorous follow-the-rules formatting). so to be able to organize my thoughts, I prefer editing the source files manually, and keep text human readable so I can easily scan then and write/correct/improve it.

thebe14: Another example from a commercial provider on why hard-breaking lines in shell examples is wrong and never done: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-lamp-amazon-linux-2.html

Try resizing the page until the content gets a narrow column. You will see that the shell examples get a horizontal scroll bar, just like ours do. Azure docs do the same: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/quick-create-cli

gwarf: I agree we shouldn't break command output, so if this exception is for this it's fine with me, especially as output is usually less useful and it's acceptable to not see it when looking at the page (ie. having horizontal scroll bar at the bottom).

Maybe we should include output only where it really makes sense as in example like https://docs.egi.eu/users/online-storage/object-storage/#access-via-s3-protocol it doesn't bring that much value.

We need to ensure that all files are styled in the same coherent way, prettier may have some choice that you may not like, but at least it provides an easy way for people to apply a common style, having plugins for various editors/IDE.

I can't agree with spaces at EOL or empty lines at EOF as I've got OCD with this, but if other agree with you we can disable the line length, there is an option for this in prettier: https://prettier.io/docs/en/options.html#prose-wrap.
I reckon that this quite simple change may make life easier for people editing files from the GitHub web interface.

content/en/users/online-storage/object-storage/_index.md
gwarf: Fine, so let's close this. Let's try to trigger again this discussion when everyone will be back from the break.

content/en/users/online-storage/object-storage/_index.md
gwarf: We are squashing PR so the details commits and last commits won't be seen. I'm/we are usually removing the minor commit messages from the final message, see https://github.com/EGI-Federation/documentation/commits/main.

If we don't have defined rules that are least enforced when validating the PR it will become a mess, no way we will disable this.

The idea was to have an automated formatter that can be used to ease people's life and help them with basic look and feel (useful for people less advanced than you on those topics) and then have automated checks when people are making commits so that we are sure to have a minimal quality for the submission. And of course both should be aligned. Using an "opinionated" formatter is also a way to end the discussions about style, as it is a sensitive topic :)

You are also free to propose a PR changing the formatter and linter settings (and the documentation/style guide).

gwarf: In fact you moved the enable line length check forward after a code excerpt that was not having any output, please break down this long command then let's try to finally merge this PR.

content/en/users/online-storage/object-storage/_index.md
thebe14: Let me update the pages with one more commit, then we can close it

PR #275, opened 2021-08-12
Shell example Copy button only copies the command

thebe14:
Improved code blocks of type "shell", "bash" and "terminal" so that the _Copy_ button only copies to the clipboard the command itself. It ignores the prompt ($) and any command output. It supports commands broken to multiple lines with backslash (\).

---

**Related issue :**


gwarf:
That looks great, thanks!
Will give it a try.

gwarf (commented):
It works quite great, the only problem I see is that for shell excerpts that contains multiple lines of commands that could be pasted all at once it will only copy the first one.
Example with https://docs.egi.eu/users/datahub/clients/#testing-oneclient-in-a-virtual-machine

With the current version everything is copied, while with the new one, only the first line is copied.

If we decide to move to have output in a separate block it would probably be nicer to have the copy command copying all lines.

thebe14:
Well, I had a one command per shell example in mind. The example you linked does not have prompt before the commands.

thebe14:
The _Copy_ button now copies multiple commands from a sample block.
For shell blocks where no line has a prompt ($), it copies everything.
For shell blocks where commands have a prompt and the command output does not, it will only copy the commands. 

> Once it encounters lines that it determines are command output, it will no longer look for commands to include in the copied text. Multiple commands, each having command output, should go into distinct shell sample blocks.

Issue #277, opened 2021-08-13
Deploy to GitHub pages workflow should not be run on forks


gwarf:
Currently when an update is pushed to the main branch of a fork the Deploy to GitHub pages workflow is triggers, and while it fails due to missing information it would be cleaner to just disable this and have this workflow run only on the EGI-Federation repository.

thebe14:
I agree

PR #280, opened 2021-08-13
Do not run deploy step for forks, fixing #277.

gwarf:
Do not run deploy step for forks, fixing #277.

Not tested, but done according to https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif and https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context.


andrea-manzi (approved):
I didn't test it, but it looks good!

PR #285, opened 2021-08-17
Added support for Windows prompt (>) in shell examples

thebe14:
This allows the Copy button in shell examples to copy only the command(s) when the code examples are for Windows and the prompt is represented by > (as opposed to $ on Linux/Mac shell examples).

---

**Related issue :**


gwarf (approved):
LGTM, thanks!

PR #287, opened 2021-08-20
Add workflow trying to deploy to alternate repo from a PR

### LINKED ISSUE #173: Provide rendering of PRs

It would be nice to have some rendering of the proposed PRs so people can take a look how the changes will look on the server before merging. 

# Summary of proposed changes

gwarf:
Add workflow trying to deploy to alternate repo from a PR.
* For security reasons only PR having the label `preview` added to them will be build and deploy for preview. See https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels, only people having at least triage access can apply/dismiss labels
* The deploy is done to an alternate repository: https://github.com/EGI-Federation/documentation-preview/
* The rendering should be available at http://docs.egi.eu/documentation-preview/{PR}
Fix #173.

### DIFF
--- a/.github/workflows/preview_pr.yml
+++ b/.github/workflows/preview_pr.yml
@@ -4,8 +4,6 @@ name: Preview pull requests
 # https://github.com/peaceiris/actions-hugo
 # Deploy to GitHub pages
 # https://github.com/peaceiris/actions-gh-pages
-#
-# FIXME: deploy to a path corresponding to the current PR
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


gwarf:
@enolfc, can you check if you think it makes sense?
I'm not that sure how the proper token may be accessible, let's try and see.
This PR may need some future PR to fix possible problems/failures.

enolfc (approved):
It looks ok, although I guess we will have to fight a bit with all the different pieces in subsequent PRs :)

gwarf:
> It looks ok, although I guess we will have to fight a bit with all the different pieces in subsequent PRs :)

Yep, I expect the same.

PR #288, opened 2021-08-20
Drop Hugo binaries and bump version used to build the documentation

### LINKED ISSUE #281: The provided hugo binary is no more used to build the site
The provided hugo binary is no more used to build the site: when moving from Travis to GitHub actions an action allowing to install a specific hugo version dynamically and making useless to store the binary in the project directory.

If we want to keep using an embed version (to save time in workflows) we need to update the workflows.
Otherwise it probably makes sense to drop the binaries directory and adjust installation notes, unless we fill that providing the binaries is useful for people willing to build/test the repository locally (making it easier).

Opinions?

gwarf:
Drop Hugo binaries and bump version used to build the documentation.
Fix #281.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       - name: Setup Hugo
         uses: peaceiris/actions-hugo@v2
         with:
-          hugo-version: "0.66.0"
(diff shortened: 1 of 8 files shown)


enolfc (approved):
👍

PR #292, opened 2021-08-23
Update PR preview (purge older files, add documentation)

gwarf:
Update PR preview (purge older files, set more options explicitly, add documentation)


.github/workflows/deploy_pr_preview.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Extract pull request preview
+        run: unzip pr.zip
+
+      - name: Deploy Pull Request preview
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          # Use GITHUB_TOKEN allowing to write to local repository
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # XXX add a way to clean deployed PR having been closed or merged
```
enolfc, line 49: I guess this comment can be removed now

enolfc (approved):
ok!

.github/workflows/deploy_pr_preview.yml
gwarf: yep, you are right. Let's see later if it works.

.github/workflows/deploy_pr_preview.yml
```diff
@@ -0,0 +1,77 @@
...
+          commit_message: ${{ github.event.head_commit.message }}
+
+      - name: Update Pull Request
+        uses: actions/github-script@v4.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            var fs = require('fs');
+            var issue_number = Number(fs.readFileSync('./PR'));
+            const output = `#### Documentation preview deployed!\`
```
enolfc, line 70:
```suggestion
            const output = `#### Documentation preview deployed!
```

enolfc: Not 100% on how js treats strings, but that seems to not belong

PR #297, opened 2021-08-24
Cosmetics

gwarf:
Various small cosmetic changes on the OneProvider documentation.

PR #299, opened 2021-08-24
Modify PR preview directory

enolfc:
# Summary

---

**Related issue :**

### DIFF
--- a/.github/workflows/deploy_pr_preview.yml
+++ b/.github/workflows/deploy_pr_preview.yml
@@ -44,7 +44,9 @@ jobs:
             fs.writeFileSync('${{github.workspace}}/pr.zip', Buffer.from(download.data));
 
       - name: Extract pull request preview
-        run: unzip pr.zip
+        run: |
+          mkdir public
+          unzip -d public pr.zip
 
       - name: Deploy Pull Request preview
         uses: peaceiris/actions-gh-pages@v3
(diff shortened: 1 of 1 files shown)


gwarf (commented):
So you mean that the zip removes the parent `public` directory that should contain the built code + the PR number?
If that's the case you will have to also adapt the part that is reading the PR number when adding the comment.

gwarf:
I'm a bit unsure, see 
Screenshot 2021-08-24 at 12 33 55

gwarf:
From #297: https://github.com/EGI-Federation/documentation/runs/3410041245?check_suite_focus=true#step:3:4

enolfc:
@gwarf can you take another look?

.github/workflows/deploy_pr_preview.yml
```diff
@@ -44,7 +44,11 @@ jobs:
             fs.writeFileSync('${{github.workspace}}/pr.zip', Buffer.from(download.data));
 
       - name: Extract pull request preview
-        run: unzip pr.zip
+        id: extract
+        run: |
+          unzip pr.zip
+          PR_NUMBER="$(cat ./PR")
```
gwarf, line 50:
```suggestion
          PR_NUMBER="$(cat ./PR)"
```

gwarf (approved):
Let's try this.

PR #301, opened 2021-08-24
Exclude files from deploy

gwarf:
Exclude files from deploy, rename workflow.


gwarf:
Let's keep this PR open and wait to confirm deploy and delete are working OK, otherwise we can make fix there.

gwarf:
Latest commit should address https://github.com/EGI-Federation/documentation/runs/3421870373?check_suite_focus=true#step:3:24

gwarf:
@enolfc can you look at this one? The `Supper-Linter` check is not working as it's renamed in this PR, thus we can ignore it.

PR #302, opened 2021-08-24
Import manual 02 and how to listed at https://docs.egi.eu/providers/operations-manuals/

gwarf:
> Work in progress

* Import manual 02: https://wiki.egi.eu/wiki/MAN02_Service_intervention_management
* Import howto 01: https://wiki.egi.eu/wiki/HOWTO01
* Import howto 02: https://wiki.egi.eu/wiki/HOWTO02
* Import how to 03: https://wiki.egi.eu/wiki/HOWTO03
* Import how to 04: https://wiki.egi.eu/wiki/HOWTO04

### DIFF
--- a/content/en/providers/operations-manuals/_index.md
+++ b/content/en/providers/operations-manuals/_index.md
@@ -28,4 +28,4 @@ documentation and procedures.
   Information required in the GOCDB when registering a site/resource centre
 - [HOWTO02 Site Certification Required Documentation](./howto02_site_certification_required_documentation)
 - [HOWTO03 Site Certification GIIS Check](./howto03_site_certificatoin_giis_check)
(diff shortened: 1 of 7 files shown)


gwarf:
@enolfc HOWTO 04 covers some part about cloud, but it's quite old, not sure what you all want to do with this.

paolini78 (approved):
thanks for fixing the mess I created. the pages look good.
We can check later on what to update in the single pages.

gwarf:
> thanks for fixing the mess I created. the pages look good.
> We can check later on what to update in the single pages.

OK, we really need to clean them, there is a lot of outdated/useless information.

As I've been able to convert some content from mediawiki to markdown, I could also add other pages that should be imported, but I guess they are not all relevant. So if you give me a list I can add them.
And some strictly grid-specific content may be better pushed at some point to the high-throughput-compute section.

PR #313, opened 2021-08-28
Updated examples for fedcloudclient returning shell commands to set env vars

thebe14:
# Summary

Updated the documentation to show how to use the new FedCloud Client (see [this PR](https://github.com/tdviet/fedcloudclient/pull/64)) with commands that return sell commands to set environment variables. This now also works on Windows, but is especially elegant in PowerShell (which can run on non-Windows platform too!).

---

**Related issue :**


content/en/users/getting-started/cli/_index.md
```diff
@@ -48,6 +47,7 @@ FedCloud client has the following modules (features):
 The FedCloud client can be installed with the `pip3` Python package manager
 (without root or aministrator privileges).
 
+<!-- markdownlint-disable no-space-in-emphasis -->
```
gwarf, line 50: Same as in #314: Instead of setting this exclusion it would be cleaner to use `$VIRTUAL_ENV/bin` instead of _$VIRTUAL_ENV/bin_.
It makes also those paths more distinguishable and I think it makes sense to use backticks for paths.

gwarf: (and you put this same change in two PRs)

thebe14: Fixed

gwarf (approved):
LGTM

PR #317, opened 2021-09-07
Update fedcloudclient command for the discovery of GPUs

sebastian-luna-valero:
# Summary

Thanks to @tdviet for sending the new command in today's Fedcloud Task Force meeting.

As discussed in https://github.com/EGI-Federation/documentation/pull/316#issuecomment-914414114 let's see what happens to line length.

PR #318, opened 2021-09-08
Try to find syntax for configuring MD013 exceptions

gwarf:
Try to find syntax for configuring MD013 exceptions, see https://github.com/EGI-Federation/documentation/pull/317/checks?check_run_id=3536232892#step:4:103 from #317.

This PR should set a max line length of 120 to be more comfortable, with line length checks disabled for code excerpts and tables that are the legit cases for longer lines.

See https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md013

### DIFF
--- a/.github/linters/.markdownlint.json
+++ b/.github/linters/.markdownlint.json
@@ -1,5 +1,9 @@
 {
-  "MD013": false,
+  "MD013": {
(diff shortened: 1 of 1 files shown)


enolfc (commented):
LGTM

PR #322, opened 2021-09-09
Build PR preview to include draft pages

### LINKED ISSUE #321: Preview of PRs does not include draft pages
The previews should also build and run so that draft pages are visible. Otherwise these previews do not make much sense for content that is added to draft pages (e.g. PR #319). Or maybe there should be no PRs for still-draft content?

# Summary of proposed changes

thebe14:
This will allow work on content that is still not public ready to be previewed.

---

Fixes #321

### DIFF
--- a/.github/workflows/build_pr_preview.yml
+++ b/.github/workflows/build_pr_preview.yml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Build and minify using Hugo
         # Build in a directory specific to the PR number
-        run: hugo --minify --destination public/${{ github.event.number }}
+        run: hugo -D --minify --destination public/${{ github.event.number }}
 
       - name: Save PR number
         run: echo ${{ github.event.number }} > ./public/PR


gwarf:
As said in #321 the goal of the preview is to have the preview of the real website, not testing some drafts, I'm not really for enabling this.

gwarf:
(and would prefer to limit the number of drafts commited to the repo)

.github/workflows/build_pr_preview.yml
```diff
@@ -57,7 +57,7 @@ jobs:
 
       - name: Build and minify using Hugo
         # Build in a directory specific to the PR number
-        run: hugo --minify --destination public/${{ github.event.number }}
+        run: hugo -D --minify --destination public/${{ github.event.number }}
```
gwarf, line 60:
```suggestion
        run: hugo --buildDrafts --minify --destination public/${{ github.event.number }}
```

gwarf: Still not willing to support this, I would prefer to have the preview really look like the final output.
And people can just commit without the draft setting and have the preview work.

PR #324, opened 2021-09-13
Lint all content

gwarf:
Always lint all files in order to avoid having files not going through the linter (or results overlooked) when changes are pushed quickly.
I've also run markdownlint on all the content to address pending issues.


thebe14 (commented):
Wondering if it really makes sense to run this duplicate checker on markdown content. It is a (sort of) valid thing to do for code, but for documentation? Especially with this aggressive checker (aggressive in the sense that flags it as a problem when only a few lines are repeated), written for code.

gwarf:
> Wondering if it really makes sense to run this duplicate checker on markdown content. It is a (sort of) valid thing to do for code, but for documentation? Especially with this aggressive checker (aggressive in the sense that flags it as a problem when only a few lines are repeated), written for code.

I think that sometimes it could help people rethink how something has been documented and try to simplify things like avoiding repeating the same steps to setup the environment before using some commands.

It's also not something that is showing that often in reports nor currently needs lots of exception, so I'm not that concerned by keeping it (At least for now).

thebe14:
In theory that sounds good. But empirically I found this:

- it sometimes flagged content that was already in the docs and I did not know what to do with it but to disable this check
- for every page where I added OS-specific steps I had to disable this check

gwarf:
> In theory that sounds good. But empirically I found this:
> 
> * it sometimes flagged content that was already in the docs and I did not know what to do with it but to disable this check
> * for every page where I added OS-specific steps I had to disable this check
> 
> The above prompted me to add instructions to the contrib guide on how to disable it.

I don't feel there is any urgency on this (especially as it's documented ;) ), so I propose to keep it for some more time, and then if we see that it mainly leads to false positive/exception we can just drop it.

enolfc (approved):
👍

PR #328, opened 2021-09-20
Import HOWTO07

gwarf:
Import HOWTO07: https://wiki.egi.eu/wiki/HOWTO07_How_to_change_the_Site-BDII.

To be reviewed/cleaned.


gwarf:
@paolini78  I feel like this one should go under HTC, we could start populating this section if you agree.

paolini78:
fine with me.

paolini78 (approved):
thanks

PR #330, opened 2021-09-20
Import How-to related to Workload Manager

gwarf:
Import How-to related to Workload Manager: https://wiki.egi.eu/wiki/HOWTO22

I feel like this documentation should just be merged with https://docs.egi.eu/users/workload-manager/

Looking at https://docs.egi.eu/users/workload-manager/#how-to-access-the-egi-workload-manager-service we may just be able to skip it, no?

### DIFF
--- a/content/en/providers/operations-manuals/howto22_accessing_workload_manager/_index.md
+++ b/content/en/providers/operations-manuals/howto22_accessing_workload_manager/_index.md
@@ -1,0 +1,354 @@
+---
+title: HOWTO22
+weight: 130
+type: "docs"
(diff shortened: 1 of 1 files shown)


enolfc:
I also prefer merging, and checking that the commands in here still work :)

paolini78:
I cannot verify the commands any more, anyway looking at https://docs.egi.eu/users/workload-manager/#how-to-access-the-egi-workload-manager-service , it seems an up-to-dated and improved version of the old HOWTO22, which we can then discard.

gwarf:
Thanks, closing this PR without merging, existing documentation is already available.

PR #334, opened 2021-09-30
Remove the RSA key markers to lazily address errors reports by gitleaks

gwarf:
Remove the RSA key markers to lazily address errors reports by gitleaks, see https://github.com/EGI-Federation/documentation/pull/333/checks?check_run_id=3755392770#step:4:1064

https://github.com/zricethezav/gitleaks

It's desirable to keep gitleaks running and showing the PRIVATE KEY tags doesn't seem to be that required/useful.


enolfc (approved):
:+1:

PR #340, opened 2021-10-14
Update VO Configuration guide

enolfc:
# Summary

Update the VO configuration guide and bring the VO entitlements mapping to the docs

---

**Related issue :**


enolfc:
Added generator of mappings here: https://github.com/EGI-Federation/fedcloud-catchall-operations/pull/137

paolini78:
if you think it would be useful having this configuration information on the VO Id Card, you can open a requirement ticket: think about the additional fields you would need, how they can be used, and let's see if it they can be implemented.

PR #375, opened 2021-12-16
Use the .html sample instead of .mlx

enolfc:
# Summary

Just make it easier to visualise the file without matlab.

---

**Related issue :**


gwarf (approved):
LGTM

PR #386, opened 2022-01-25
Add GGUS to exceptions

Thysk:
GGUS is behind a certificate, and therefore cannot be accessed by the link checker

This is required for Rucio documentation to not have link exceptions added


gwarf:
Thanks! :)

PR #417, opened 2022-02-21
Create release-schedule.md

ggrein:
# Summary

GGUS Release Schedule

---

**Related issue :**


gwarf:
@ggrein seems my email came late, but I'm not at all sure about the relevance of this for this doc site, if we include this info here we would need to do it for all central services, and I don't think it's planned nor that we have it.
Can't it be moved to GGUS release_notes page itself?

ggrein:
Why not?
I think this is a good idea.

gwarf:
I'm currently not sure what's the purpose of this page, and I don't think we agreed/need to collect this info for other services.
docs.egi.eu is more meant to cover documentation about the services, mainly for people to use them, and I don't (yet?) see the benefit of moving this page here.
If you really would like to share this kind of schedule maybe it could be easier/cleaner to have it with https://ggus.eu/index.php?mode=release_notes ?
It would also be easier for you to update it when releasing a new version.

gwarf:
Closing this PR, it was agreed to put this in the release_notes page of GGUS.

PR #423, opened 2022-03-07
Redirect pages for old users documentation page locations

### LINKED ISSUE #420: Add redirects from the old documentation sections
Create HTML pages in each (top) folder of the old documentation structure, that will redirect to the new documentation pages. These do not show up in the documentation left-side navigation bar, and we can keep them (at least for a while) as a safety net.

thebe14:
# Summary

Links to user documentation in older documents/material are now incorrect, after the docs structure change. Redirect to the current page from old locations.

---

Fixes #420


gwarf:
Thanks looks good.
Apparently you didn't create the PR from an up to date main branch, but once we squash the PR it should be fine, only the new files are changed.
The linter reports the missing `<title>` in the `<head>` elements, could you add them?

gwarf:
I'm not sure about the title:     `<title>404 Page not found</title>`: fist the page will return an HTTP code 200 and it's more a redirection than a page not found.
Maybe you could replace by something more generic like `<title>Redirecting to new location</title>` ?

content/en/users/check-in/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>404 Page not found</title>
+    <link rel="stylesheet" type="text/css" href="/scss/main.css">
```
gwarf, line 5: Ideally you shouldn't use an hardcoded path like here, as it will fail on things like the test instance.

content/en/users/check-in/index.html
```diff
@@ -0,0 +1,15 @@
...
+    <link rel="stylesheet" type="text/css" href="/css/error.css">
+    <meta http-equiv="refresh" content="0; url='../aai/check-in/'" />
+  </head>
+  <body class="error">
+    <p><img src="https://egi.eu/wp-content/uploads/2016/05/cropped-logo_site-1-300x300.png" alt="Logo" style="height: 100px;"></p>
```
gwarf, line 10: It would be preferable to use an included image than relying on one in the wodpress site (that will get removed soon).

gwarf: We should already have a logo in the public dir.

gwarf:
Have you bee looking at some point at https://gohugo.io/content-management/urls/#aliases?
I've also seen things like https://www.scripteverything.com/hugo-redirect-301-layout/.

But I'm totally fine with merging this PR for those pages to address this as quickly as possible, it's just that it could be interesting as an alternate way later.

content/en/users/check-in/index.html
gwarf: (this is something not ideal but I would be OK for merging despite this)

content/en/users/check-in/index.html
gwarf: I think it's the main point to be fixed before being able to merge.

gwarf (changes requested):
Thanks, looks great.
I've added a few comments, please review them. I would personally be OK for merging after having updated the title and used a local logo file.

thebe14:
> Have you bee looking at some point at https://gohugo.io/content-management/urls/#aliases? I've also seen things like https://www.scripteverything.com/hugo-redirect-301-layout/.

Yes, it does not add anything. More complicated to setup and maintain. Also, these are temporary and we'll get rid of them in a relatively short period.

content/en/users/check-in/index.html
thebe14: The docu landing page also links to that. Will include it locally.

enolfc (approved):
LGTM if there are no broken links found :) Agree with @gwarf comments, but also need to merge asap

thebe14:
> I'm not sure about the title: `<title>404 Page not found</title>`: fist the page will return an HTTP code 200 and it's more a redirection than a page not found. Maybe you could replace by something more generic like `<title>Redirecting to new location</title>` ?

Replacing titles....

thebe14:
> Thanks looks good. Apparently you didn't create the PR from an up to date main branch, but once we squash the PR it should be fine, only the new files are changed. The linter reports the missing `<title>` in the `<head>` elements, could you add them?

But I did! First I did a rebase of my main branch from the EGI main one, took me half a day to merge the damn thing (also why I didn't push a PR on Friday). Then created a new branch from that, added the content I prepared Friday, and created the PR.

thebe14:
Do not merge yet, let me replace with aliases. They seem to work okay.

content/en/users/check-in/index.html
gwarf: I was meaning the deployed preview, the PR is prepended in the URL path, see https://docs.egi.eu/documentation/423/users/check-in. Thus any `/something` will break.
If there is no simple solution it's acceptable for such a page.

thebe14:
@enolfc @gwarf The remaining dead links are GitHub links (that are not publicly accessible?)

gwarf:
> @enolfc @gwarf The remaining dead links are GitHub links (that are not publicly accessible?)

No, what I currently see is an error reaching an ops portal URL: https://github.com/EGI-Federation/documentation/runs/5460594664?check_suite_focus=true#step:4:228

gwarf (approved):

PR #427, opened 2022-03-22
Update authors file.

gwarf:
Update authors file.
Use GitHub account when no email is provided.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,7 +5,9 @@ Baptiste Grenier <baptiste.grenier@egi.eu>
 
 Contributors
 ------------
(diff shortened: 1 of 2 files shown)


gwarf:
@enolfc I wonder if we should just link to the GitHub profiles, letting users manage personal data (name, email) as they want.

gwarf:
OK so we just drop the full `AUTHORS` file and don't document anything in replacement?

gwarf:
So I've kept the info about repo maintainers, I guess it can be useful if people want to be able to contact someone directly.

PR #429, opened 2022-03-25
Add files via upload

EGI-ILM:
I think the documentation has not been updated to reflect that now it says "Identity" instead of "Account". I made the changes in two different Pull requests. Perhaps the old image for account has to be deleted too.


gwarf:
Can you please move this to #430, as it's the PR that is starting to use this image? It's better to have all this in a single/atomic PR. Would be good to also clean the unused image.

EGI-ILM:
Done!  You are right, I used the GUI and was not the best way

PR #458, opened 2022-06-09
Update code owners

gwarf:
Review and update code owners according to current status of the repo, as apparently some PR for Check-in are not properly automatically assigned to Valeria.
Still relying on individual users more than on teams.


enolfc (commented):
I see missing:
`content/en/users/compute/content-distribution/` maybe for @CatalinCondurache 
`content/en/users/tutorials/` for @thebe14 ?

On the data side, do we want to be more specific? @andrea-manzi ?

enolfc (approved):
LGTM

PR #471, opened 2022-07-07
Fix typo in json

enolfc:
# Summary

Just add a missing comma

---

**Related issue :**

### DIFF
--- a/content/en/providers/cloud-compute/openstack/aai.md
+++ b/content/en/providers/cloud-compute/openstack/aai.md
@@ -757,7 +757,7 @@ the URLs on your configuration:
    {
      "type": "HTTP_OIDC_ISS",
      "any_one_of": [
-       "https://aai.egi.eu/auth/realms/egi"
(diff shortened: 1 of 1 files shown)


gwarf (approved):
LGTM

gwarf:
I feel the link checker is a bit confused with the way the aai page is created (ie having a subdir with a `_index.md` vs having here an `aai.md`). I don't think we can do much for now, it would either need to move to subdir + _index.md (which in theory shouldn't be needed) or trying to fix the link checker...

gwarf:
Nice try but I feel like in that case the link is just ignored by the link checker and not tested at all. I'm not fond of those hugo specificities when they are not strictly needed and not compatible with usual markdown. And as an example will make link that don't work when looking at the  source code in GitHub. I'm also not sure how it would work with the preview, I suspect it may break.
As/if it doesn't allow to have the test to success for real, I tend to prefer keeping the proper link not passing the test, hoping that at some point we find a clean way to handle this.

PR #475, opened 2022-07-13
Document certificate provisioning

gwarf:
Document certificate software provisioning for central services. I've documented this under the Collaboration Tools internal service, the SDTP is being updated and discussions are still to take place on the exact list of components. But until know, we can keep it here, unless anyone is having a better suggestion.

Documentation status:
- [X] issuing host certificates
- [x] issuing robot certificates


gwarf:
See https://docs.egi.eu/documentation/475/internal/collaboration-tools/certificates/

PR #477, opened 2022-07-22
[Check-in] Add OP migration section in the end-user guide

nikosev:
# Summary

This PR introduce the following changes:

- Add a section for the OP migration in the end-user guide
- Add note for the minimum required version of oidc-agent

### DIFF
--- a/content/en/users/aai/check-in/obtaining-tokens/token-portal/_index.md
+++ b/content/en/users/aai/check-in/obtaining-tokens/token-portal/_index.md
@@ -40,3 +40,18 @@ follow the steps below:
    ![EGI Check-in Token Portal Refresh Token](check-in-token-refresh-token.png)
    1. The value of the Refresh Token
    2. The command to generate new Access Token using the Refresh Token
+
+## EGI Check-in OpenID Provider Migration
+
+Currently we are migrating EGI Check-in OpenID Provider to a new Authorization
+Server, as result the Refresh Tokens, that was issued by the legacy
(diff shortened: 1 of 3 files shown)


content/en/users/aai/check-in/obtaining-tokens/token-portal/_index.md
```diff
@@ -40,3 +40,18 @@ follow the steps below:
...
+
+## EGI Check-in OpenID Provider Migration
+
+Currently we are migrating EGI Check-in OpenID Provider to a new Authorization
+Server, as result the Refresh Tokens, that was issued by the legacy
+Authorization Server, will not be accepted by the new one and you won't be able
+to generate new Access Tokens. In this case you will need to create a new
+Refresh Token using the dedicated instance of EGI Check-in Token Portal, which
+is integrated with the new EGI Check-in OpenID Provider. In the following table
+you can find the URLs of each instance of EGI Check-in Token Portal:
```
gwarf, line 52:
```suggestion
As of July 2022, we are migrating EGI Check-in OpenID Provider to a new Authorisation
Server, as a result the Refresh Tokens, issued by the legacy Authorisation Server,
will not be accepted by the new one and you won't be able to generate new Access
Tokens with them. In this case you need to create a new Refresh Token using the
environment-specific EGI Check-in Token Portal, which is integrated with the new
EGI Check-in OpenID Provider. In the following table you can find the URLs of the
various instances:
```

vardizzo-lab (approved):
OK. Approved

PR #499, opened 2022-09-02
Reorganise pages under the providers section, as requested by Gergely.

gwarf:
Reorganise pages under the providers section, as requested by Gergely.

Getting rid of the Getting started section, the joining section is moved to the top level, including content for integrating service or IdP with Check-in.
Other parts of the getting started are moved next to the other operational documentation (manuals, how-tos.,..).


paolini78 (approved):
thanks

paolini78 (approved):
thanks

paolini78 (approved):
thanks

PR #511, opened 2022-10-14
Add ignore file for gitleaks, but disable gitlekas as ignore file is not working yet

gwarf:
Only look at current tree, not at git history


enolfc (approved):

gwarf:
I wasn't yet able to find a way to make this work, but I'm not yet sure if it's due to our env, or just a broken/buggy feature.
For the record I've asked some feedback in a (unfortunately closed) issue: https://github.com/github/super-linter/issues/3407

We may want to test/debug super-linter locally, to see what it does.

gwarf:
@enolfc the gitleaks exclusion file not being working for now, and this check being breaking all PR for now, should we just disable it for some time?

enolfc:
> @enolfc the gitleaks exclusion file not being working for now, and this check being breaking all PR for now, should we just disable it for some time?

Ok, let's disable it.
But we should remind ourselves this needs to be fixed in the future

PR #512, opened 2022-10-17
Remove links to EGI Marketplace

sebastian-luna-valero:
# Summary

---

**Related issue :**

### DIFF
--- a/content/en/users/getting-started/_index.md
+++ b/content/en/users/getting-started/_index.md
@@ -163,9 +163,6 @@ the users' requests before granting the access to the resources.
 2. Enroll the `vo.access.egi.eu` Virtual Organisation by following the
    [enrollment URL](https://aai.egi.eu/registry/co_petitions/start/coef:240).
    Make sure you use your EGI Check-In account for the enrollment.
-3. Browse EGI services in the
-   [EOSC Marketplace](https://marketplace.eosc-portal.eu/providers/egi-fed)
-   and place an order.
 
 The grant to run applications is initially valid for 6 months and can be
 extended/renewed upon request. These resources are delivered
--- a/content/en/users/getting-started/_index.md
+++ b/content/en/users/getting-started/_index.md
@@ -81,9 +81,7 @@ of the concepts that act as building blocks for the AAI implementation.
 ## Requesting resources
 
 Depending on the access conditions, a service (or an instance of the service)
-may be open for any user, or it may require requesting access (ordering). The
-[EGI site](https://www.egi.eu/services/) together with the connected
-[EGI Marketplace](https://marketplace.egi.eu) streamlines the ordering process.
+may be open for any user, or it may require requesting access (ordering).
 
 EGI services use the following types of access conditions:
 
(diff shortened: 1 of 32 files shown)


content/en/users/getting-started/_index.md
```diff
@@ -161,30 +159,13 @@ the users' requests before granting the access to the resources.
...
-
-Once granted access, each user will have a grant with a predefined quota of
-resources, which can be used to run the application of choice. This grant
-includes:
-
-- up to 4 CPU cores,
-- 8 GB of RAM,
-- 100GB of block storage.
+1. Create an [EGI Check-In account](../aai/check-in/).
+2. Enroll the `vo.access.egi.eu` Virtual Organisation by following the
```
enolfc, line 163: Should this order be shuffled: first order in EOSC and then request membership? This would facilitate identifying the people that requests access, no?

sebastian-luna-valero: I just did a test. To order in EOSC you need to authenticate. Maybe following the current order they first get an Check-In account that they can use to order the service later on in the EOSC Marketplace. Would this simplify things? Happy to change the order as proposed if that's more helpful.

gwarf: In theory a customer could authenticate to the EOSC Portal AAI using their own existing institutions IdP connected to the EOSC Portal AAI and possibly totally outside of Check-in, it wouldn't allow to register to a Check-in and grant access to our resources, but it's possible it could happen.

Maybe I miss something, but should we really force people to create a Check-in account before placing an order in the MP? In fact what's the added value for them to use the MP instead of just creating a Check-in account and enrolling directly to the VO?
What's the value of opening an MP request once you are already in the vo.access.egi.eu VO?

sebastian-luna-valero: > In theory a customer could authenticate to the EOSC Portal AAI using their own existing institutions IdP

> should we really force people to create a Check-in account before placing an order in the MP

Correct. I was just trying to simplify the process as a whole, using just the Check-In account. For outsiders jumping from one system to other might be confusing.

> what's the added value for them to use the MP instead of just creating a Check-in account and enrolling directly to the VO?

>  What's the value of opening an MP request once you are already in the vo.access.egi.eu VO?

No added value for the end user. Going through the MP first would allow EGI to quantify the interest in their services out there.

---

Having said all of that, I think what's important in this PR is to remove the links to the retired EGI marketplace. I am happy to adapt the suggestions above as soon as we agree on something. I am flexible with the options discussed.

enolfc: I agree the main objective of the PR is removing links to the MP

To be pragmatic we may just remove the step 3 below. This section is about getting access to the pilot allocation which is just vo.access.egi.eu. The EGI marketplace was in this section becase we used to approve automatically access to the VO via the order, but that's no longer the case, so no need to artificially add the EOSC marketplace if it's not bringing value to us or the user

gwarf (approved):
LTGM, thanks!

PR #526, opened 2022-10-21
Bump hugo version to latest release

gwarf:
Bump hugo version to latest release: https://github.com/gohugoio/hugo/releases/tag/v0.104.3

PR #527, opened 2022-10-24
GitHub actions: use updated octokit call

gwarf:
GitHub actions: use updated octokit call.
This should fix issues with deploying pull request previews, as identified in #526 .
Tested in https://github.com/gwarf/documentation/actions/runs/3312681659/jobs/5469654742

### DIFF
--- a/.github/workflows/deploy_pr_preview.yml
+++ b/.github/workflows/deploy_pr_preview.yml
@@ -26,7 +26,7 @@ jobs:
         uses: actions/github-script@v6.3.3
         with:
           script: |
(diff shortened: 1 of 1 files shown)


enolfc (approved):
LGTM! It build so I guess we are happy :)

gwarf:
I think that one of the problem with workflow files, is that it's the version from the main branch that is used, so we don't see yet the effect of them.

gwarf:
It worked: https://github.com/EGI-Federation/documentation/pull/526#issuecomment-1289145377 🎉

PR #530, opened 2022-10-25
Update docs about deleting downtimes starting in 24 hours or less

gregcorbett:
# Summary

This PR will update the docs about deleting GOCDB downtimes starting in less than 24 hours or less. Being able to delete such downtimes was a bug that we fixed in GOCDB 5.9.0: https://github.com/GOCDB/gocdb/pull/279

PR #531, opened 2022-10-25
Disable markdownling check for ops portal URL protected by Check-in

gwarf:
Disable markdownling check for ops portal URL protected by Check-in


gwarf:
This should fix the issue reported in https://github.com/EGI-Federation/documentation/pull/530

enolfc (approved):
👍

PR #244, opened 2021-07-09
Updated docsy theme to latest and fixed all styling issues

thebe14:
# Summary

- Updated the docsy theme to latest version
- Reduced font size to make the documentation usable at 100% zoom level
- Added support for tabbed content on pages (e.g. for installation steps that are different on Linux and Windows)
- Added copy to clipboard to code blocks
- Fixed last updated section to show user's name (with a mailto link if email available)
- Fixed double hr at bottom of pages without child pages
- Added icons to top menu
- Added repo stats to footer
- Fixed landing page feature blocks

PR #541, opened 2022-11-09
Missing };

EGI-ILM:
I have not read the code, just placed it according to the indentation. Please VERIFY that this is located in the right place.


static/js/egi.js
```diff
@@ -65,6 +65,7 @@ jQuery(document).ready(function () {
           $('.progress').stop(true, false, true).fadeToggle(100);
       });
   }
+};
```
gwarf, line 68:
```suggestion
});
```
Apparently there is also a missing `)`.
That said I'm not sure what's the exact purpose of those changes, if they are still required, and what may not be correctly working if the current version is broken.

@thebe14 I think you are the one that added this, any thoughts?

gwarf:
@enolfc , @andrea-manzi what do you think of this one?

andrea-manzi:
can we deploy and see what is changing?

gwarf:
> can we deploy and see what is changing?

I've just enabled the preview.
From my local tests, it seemed that the change I proposed was necessary to resolve warnings/errors from the dev tools.
Didn't saw much changes in the UI, but wasn't sure what may be broken or not.

static/js/egi.js
thebe14: There is definitely missing `});` although I am not clear how I could commit that and still have the things work.

thebe14:
Commit it with the `});` added, I will start working on the documentation next week, will check that the category icon styles are properly applied in the sidebar.

gwarf:
OK, thanks Levente, will merge this once the tests are green.
You purposely disabled linting those files in #244, but it may have caught this ;)

gwarf (approved):
LGTM

PR #558, opened 2022-12-02
Ssh OIDC tokens

marcvs:
I found that refresh-token endpoint again...

I think it's a really bad idea to expose that to users.

I've updated the two "alternatives" and added a third one.

"Alternatives" is in "", because I think that the option of spreading refresh tokens to the infrastructure (i.e. <https://aai.egi.eu/token>) should be the _undocumented_ alternative.


gwarf:
@enolfc, @vardizzo-lab and @NicolasLiampotis please take into account the point made by Marcus regarding the https://aai.egi.eu/token endpoint.

marcvs:
> 

Rotating refresh tokens would allow users to notice if the RT was stolen: If somebody else uses it, the own RT won't work any longer.  The problem is, that users need to learn to trigger some kind of incident response in such cases...

NicolasLiampotis:
> The problem is, that users need to learn to trigger some kind of incident response in such cases...

The malicious client’s refresh token (or any subsequently issued refresh tokens) will automatically get revoked when the legitimate client tries to use it so it's not really necessary to trigger an incident response process.

marcvs:
> > The problem is, that users need to learn to trigger some kind of incident response in such cases...
> 
> The malicious client’s refresh token (or any subsequently issued refresh tokens) will automatically get revoked when the legitimate client tries to use it so it's not really necessary to trigger an incident response process.

Perfect! So: I'm all for refresh-token rotation.

gwarf:
If one discovers that a token got stollen and used, I guess the security team would really want to be aware and be able to look into this, understanding what happened, the exact impact, if protections could be put in place and so on.
We may at least want to document (one the token app page? in docs.egi.eu? in both?) that users should report this (to whom and how).

enolfc (approved):
This looks good to me.

gwarf:
@marcvs , can you please rebase the branch? Thanks!

marcvs:
> @marcvs , can you please rebase the branch? Thanks!

Done

PR #570, opened 2023-01-30
Theme improvements

### LINKED ISSUE #5: Add EGI branding to the theme
Add EGI branding to the theme.

- @dscardaci suggested to remove the list of sections/subsections at the bottom of pages

### LINKED ISSUE #249: Font size too small after theme update

Font size too small

thebe14:
# Addressing remaining open issues to the docs theme

- Updated the EGI logo in the top navbar, fixes #5 
- Increased the theme font some more, fixes #249 
- Fixed the **Copy** button on the code/shell examples. It no longer has a double shadow, is now more opaque so that it works well when overlaps text.


assets/scss/_variables_project.scss
```diff
@@ -10,7 +10,7 @@ $primary: rgb(80, 135, 180) !default;
 
 // Fonts
 
-$font-size-base: 0.8rem !default;
```
gwarf, line 13: 🍾 :)

thebe14: Yeah, but on my laptop with very high resolution on a smallish screen, I have to use scaling. Both on my laptop screen (250%) and on the external monitor connected to it (125%). And this docs site is unusable, unless zoom is at 70% or less. Especially the left sidebar (created issue to make it resizable at least).

gwarf:
The white logo is certainly more readable, but it's not an official one, and not aligned to https://confluence.egi.eu/display/EGIG/Use+of+EGI%2C+EGI+Foundation+and+EGI+Federation.

But I guess with the issue with the background in blue, this would require a general update of the look and feel, so probably something that can be postponed to later. (maybe also once we will have adopted the next/future bootstrap 5 based docsytheme, cf https://github.com/google/docsy/discussions/1359)

So fine with me to merge this version, and then review the look and feel according to the guidelines we may/will get.

gwarf (approved):
LGTM

thebe14:
Exactly. Let's also close the linked issues, we can revisit them later if needed.

thebe14:
I am done

PR #572, opened 2023-02-02
Remove fedcloudcli env var EGI_SITE

thebe14:
# Summary

The FedCloud CLI was using an env var EGI_SITE to set the site on which to operate. This is not working since a while now, because of a bug in a Python lib. Details at https://github.com/tdviet/fedcloudclient/issues/150. 

Thus I removed this from our user docs, will update when a workaround is implemented in a new version of FedCloud CLI.


sebastian-luna-valero (approved):
Thanks!

enolfc (approved):
LGTM

gwarf (approved):
LGTM, I assume all this is properly working currently.

PR #577, opened 2023-02-20
Add steps to configure MinIO with Check-in

sebastian-luna-valero:
# Summary

---

**Related issue :**

### DIFF
--- a/content/en/users/tutorials/minio-oidc/_index.md
+++ b/content/en/users/tutorials/minio-oidc/_index.md
@@ -1,0 +1,157 @@
+---
+title: "Access MinIO with EGI Check-in Virtual Organisations"
+linkTitle: "MinIO and Check-in"
+type: docs
+weight: 170
+description: >
+  Configure access to MinIO console with EGI Check-in Virtual Organisations.
+---
+
+## Overview
+
+This tutorial describes how to configure access to 
+[MinIO console](https://min.io/docs/minio/linux/operations/external-iam.html)
+using [EGI Check-in](../../aai/check-in) as external OpenID Connect Identity
+Provider.
+
+## Prerequisites
+
+Be familiar with the steps to configure an OpenID Connect Service Provider with
+[EGI Check-in](../../../providers/check-in/sp/#openid-connect-service-provider).
+
+This tutorial assumes that:
+
+- You have a valid EGI ID (account), you can
+  [sign up here](../../aai/check-in/signup).
+- You are a member of a
+  [Virtual Organisation](../../aai/check-in/joining-virtual-organisation).
+- You have deployed MinIO and have access to the console as `admin`.
+
+## Step 1: Get the OIDC entitlement for the Virtual Organisation
+
+See the
+[page about getting tokens from Check-in](../../aai/check-in/obtaining-tokens/token-portal/).
+You will get a `curl` command to get your OIDC entitlements.
+
+Select the entitlement for the Virtual Organisation that you want to
+enable access to. For example, here is the entitlement for the
+`vo.access.egi.eu` Virtual Organisation:
+
+```
+urn:mace:egi.eu:group:vo.access.egi.eu:role=member#aai.egi.eu
+```
+
+## Step 2: Configure a new policy in MinIO console
+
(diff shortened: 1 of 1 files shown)


andrea-manzi:
@sebastian-luna-valero one first comment, is this a user tutorial or should be more of a Service provider tutorial? maybe we can start adding tutorials also under the service provider section

sebastian-luna-valero:
Hi @andrea-manzi  
Happy create a Service provider tutorial and move it down there.
@gwarf please let us know your thoughts.
Thanks!

gwarf:
> Hi @andrea-manzi Happy create a Service provider tutorial and move it down there. @gwarf please let us know your thoughts. Thanks!

So I'm unsure about who is the exact target, how this works, maybe it's a cross border case or a sign that our users/providers separation is sometimes too manichaeistic.
Currently it indeeds look like it's for service providers, that said if compute users can also/usually work as some kind of service providers for their own community, by deploying services on top of our resources...
I guess we could add it under the part for service providers, and link it in some way from the cloud compute part, if it's also something that can be relevant for a given user/community/VO willing/able to deploy it in a VM?
@enolfc any opinion here?

enolfc:
From a purely "Cloud Compute" point of view, this is user space and I would keep it under users. Using Check-in is great, **but** ToU and Privacy Notice need to be in place so this may not be so straightforward to setup for users.

@andrea-manzi do you think this is a viable option for providers to deliver online storage?

andrea-manzi:
> From a purely "Cloud Compute" point of view, this is user space and I would keep it under users. Using Check-in is great, **but** ToU and Privacy Notice need to be in place so this may not be so straightforward to setup for users.
> 
> @andrea-manzi do you think this is a viable option for providers to deliver online storage? 
> 
> 

Yes i think so, this I why we may include it in the service providers doc. For cloud providers already integrated with check-in there should be no problem to provide the ToU and Privacy Policy.

enolfc:
What we do about this one? 
Re-reading the text, there is no mention on how MinIO is to be deployed, so it's not a very user-oriented tutorial. If we agree this is an option to offer "Online Storage" service, it makes sense to have it in the providers section of the docs. Now:

1. do we try to mimic User docs organisation in providers and add this under Data -> Online Storage? If we do this, and we reorganise the rest of the content it would break links considerably

1. do we just add another top level section "Online Storage"? 

1. do we add it in the current HOWTOs section?

I don't have a clear preference on this. Maybe option (2) is not breaking things and making it easy to find the content

gwarf:
> What we do about this one? Re-reading the text, there is no mention on how MinIO is to be deployed, so it's not a very user-oriented tutorial. If we agree this is an option to offer "Online Storage" service, it makes sense to have it in the providers section of the docs. Now:
> 
> 1. do we try to mimic User docs organisation in providers and add this under Data -> Online Storage? If we do this, and we reorganise the rest of the content it would break links considerably
> 2. do we just add another top level section "Online Storage"?
> 3. do we add it in the current HOWTOs section?
> 
> I don't have a clear preference on this. Maybe option (2) is not breaking things and making it easy to find the content

For 1, we can add alias to limit link breakage, but I'm not sure it's currently required.
I would avoid adding them to the Howto section, and in fact I would ideally see the content from that section moved to other better/proper places like many under HTC (or even dropped once they will be no more relevant).

So I guess that for now the step 2 makes sense as it is the easiest and more reasonable considering the time/effort available.
It doesn't prevent doing a reorganisation after, keeping the old links working using aliases.

andrea-manzi:
Yes i agree we can add a new section as suggested in point 2. To be further extended in the future with more documentation for storage installation, but we could start from what Sebastian added for Minio. @sebastian-luna-valero do you think you can apply the changes ?

sebastian-luna-valero:
Sure, it's now done.

enolfc:
I think you need to add also a `online-storage/_index.md` for this to be rendered. See the datahub dir

sebastian-luna-valero:
true, thanks!

sebastian-luna-valero:
ok, all tests passing now

andrea-manzi (approved):
great thanks!

gwarf (approved):
LGTM, thanks!

PR #578, opened 2023-02-21
Fix syntax error in RADL file

gwarf:
Fix syntax error in RADL file, I wasn't able to deploy a VM using the code from the example.


enolfc (approved):
Thanks for spotting this!

PR #584, opened 2023-03-06
Import a GitHub action to check spell

gwarf:
Import a GitHub action to check spell: https://github.com/rojopolis/spellcheck-github-actions

Re-using configuration from https://github.com/GOCDB/GOCDB.github.io/blob/main/.github/workflows/validate.yml


enolfc:
Is there any way to ignore some words?

gwarf:
> Is there any way to ignore some words?

Yes it's possible to add a whitelist, like done here: https://github.com/GOCDB/GOCDB.github.io/blob/main/.github/workflows/config/spellcheck-permitted-words.txt

With the generated output file, we could likely easily come on with a basic list of permitted words. And see if we have that many errors.

I'm also using this PR to test another extension, that is having more stars and uses various tools relating to language: https://github.com/marketplace/actions/check-spelling
Let's see which ones works the best for us.

gwarf:
For the record, the other tool is reporting/commenting directly in my fork: https://github.com/gwarf/documentation/actions/runs/4344576094

gwarf:
@enolfc, I've removed the initial spell action in favour of https://github.com/marketplace/actions/check-spelling that is much more complete.
It's currently running via the event  [pull_request_target](https://github.com/check-spelling/check-spelling/wiki/Configuration%3A-Workflows#pull_request_target) that is expected to be more secure and working with PR from fork/clones by people not in the organisation.
I'm not used to this one, so we will have to see how it works and potentially have to adapt the event filtering, in case it's not properly triggered.
The plugin author provides various documentation on running via different events: https://github.com/check-spelling/check-spelling/wiki/Configuration%3A-Workflows

I've addressed all the errors reported, and added some words to allowed lists, it would benefit from some cleaning (like moving some words to different type of exclusion list, cf https://github.com/check-spelling/check-spelling/wiki/Configuration-information) but I don't have time for this right now.

enolfc (approved):
LGTM. I hope it's not too of a pain to deal with it ;)

gwarf:
> LGTM. I hope it's not too of a pain to deal with it ;)

I think I went through the main part, reviewing and addressing the reports from all the existing content.
Later we may see if we can/should switch to checking only the changed files.
But let's first see how it goes, we can revise using this if needed, but it was able to identify many typos or things that were better to be fixed.

PR #586, opened 2023-03-08
Lint and spell only changed files to limit wast of resources.

gwarf:
Lint and spell only changed files to limit wast of resources..

The codebase should already be in a sane situation, lint and spell checks have been running across all the codebase.


enolfc (approved):
LGMT once the linting issues are fixed

PR #587, opened 2023-03-09
spelling: use_sarif is incompatible with only_check_changed_files

gwarf:
The option use_sarif is incompatible with only_check_changed_files.

This option is still quite new, see https://github.com/check-spelling/check-spelling/wiki/Feature:-Sarif-output


enolfc (commented):
LGTM! why is the warning still there?

gwarf:
> LGTM! why is the warning still there?

Not sure if it relates to how pull_request_target event works in this context or to something else :)

PR #589, opened 2023-03-10
Import .editorconfig

gwarf:
* Import .editorconfig, see https://editorconfig.org.
* Currently only general settings are available, but they can be split depending on the file extension/type.
* Fix input name in check-spelling action


enolfc (approved):
:+1: I was not aware of this editorconfig. Need to review my (very simple) vim config

PR #593, opened 2023-04-05
[Check-in] Introduce new attributes to replace the deprecated ones

nikosev:
# Summary

This PR introduces the following changes:

- Add `distinguishedName` in the deprecated attributes table of the SAML migration guide
- Remove `eduPersonScopedAffiliation` attribute from "Affiliation" attribute description
- Add description for "Subject DN" attribute
- Add description for "Issuer DN" attribute
- Update Claims/Scopes table
- Update Attribute Descriptions


gwarf:
So it seems that due to #589 we invited a new checker to join the party, welcome https://github.com/editorconfig-checker/editorconfig-checker :)
That said I think we may have to just disable it for markdown files, or at least disabling the part on indentation check, as what is currently reported as error, seems legitimate :/
So we can likely ignore it for this PR, we will address this conf in a separate PR.

gwarf:
I've opened #594 that should help.

vardizzo-lab (approved):
The PR is approved.

PR #594, opened 2023-04-07
Prevent editorcondig-checker to look in markdown files

gwarf:
Prevent editorcondig-checker to look in markdown files as it triggers too many false positives, cf #593 .


enolfc (approved):
👍

PR #598, opened 2023-05-08
Update tutorials

sebastian-luna-valero:
# Summary

As requested via: https://docs.google.com/document/d/1nN3vw0GasnnMlM711se62Shfc3LHLP90CNh82lPmqds/

---

**Related issue :**


sebastian-luna-valero:
Unrecognized words (3)

```bash
cnica
Instrumentaci
ncia
```

Correspond to:

```
Politècnica
Instrumentación
València
```

which I added to `allow.txt`. How can I solve this, @gwarf ?

sebastian-luna-valero:
Solving spelling issues brings up loads of markdown errors. Indeed I chose html since markdown tables are not appropriate for tables like this. I would appreciate input on the best way forward.

sebastian-luna-valero:
Thanks for your feedback @enolfc 

Let's also wait for @glarocca 's feedback.

gwarf:
I find the page quite/too long to load, and hard to navigate. It also basically hides the other tutorials.
I wonder if it wouldn't be best to have the parent page as a kind of extended table of content, allowing people to easily/quickly identify interesting things for them, and then have another page (like video training modules or something "attractive/clear enough") with the videos.
Personally when I'm looking for some documentation/tutorial I definitely look at videos only if I don't have any other choice, and the first content I look for is things where I can read as I need/want and use copy and paste.

Adding raw html code is always causing issues with the automated lint, I guess the solutions would be either to exclude the problematic checks for the HTML code, or to use some (custom) short codes and/or partial template (there is one for youtube in fact: https://gohugo.io/content-management/shortcodes/#youtube):
- https://gohugo.io/content-management/shortcodes/
- https://gohugo.io/templates/partials/

For the spelling issue with accentuated later, it would need some research, in the meantime we can bypass this if needed, but it would be nice to have an issue opened for tracking this later.

EGI-ILM:
I agree that a quick "index" to identify things might look better.
Maybe even the videos can be organised in sort of "Courses" or "learning venues", like:

- **EGI**, with EGI 101, the Federation, etc
- **Cloud Compute**, containing videos and tutorials about creating a VM, etc
- And so on...

glarocca (commented):
It looks good to me. Thx

sebastian-luna-valero:
Note: Using raw HTML was the only solution I found to work with big tables in an easier way. Unless there is a better alternative I would propose to exclude it from the linting.

gwarf:
> Note: Using raw HTML was the only solution I found to work with big tables in an easier way. Unless there is a better alternative I would propose to exclude it from the linting.

I've already mentioned other options, with the short codes and partial templates.
If you don't want or can't look into this, please add marker exclusion around your HTML stuff.

https://docs.egi.eu/documentation/598/users/tutorials/ is now clearly easier to read, there is no mention of ad-hoc subsection, is it on purpose?
I wouldn't be against having a quick introductory sentence referring to it, after the 4 "skill levels".

sebastian-luna-valero:
Hi All,

I discussed with @glarocca and we think this is ready to be merged.

Does anybody have any other proposal? or can we go ahead?

Best regards,
Sebastian

enolfc (approved):
LGTM

EGI-ILM:
I think it looks better. My only comment would be to have some sort of index or similar. For example, in the "Intermediate" page I have to scroll quite a bit to navigate the contents, there is no TOC or similar element that can tell me in a single flash what sort of things I can expect.

sebastian-luna-valero:
Thanks for the suggestion @EGI-ILM 

but I would appreciate help; does anybody know how to create a dynamic TOC from this HTML table?

or should I manually create a static TOC instead?

I don't think this helps? https://gohugo.io/content-management/toc/

PR #600, opened 2023-06-01
Fix relative links in adhoc tutorial to create first VM

sebastian-luna-valero:
# Summary

Strange that this wasn't picked up in https://github.com/EGI-Federation/documentation/pull/598

---

**Related issue :**

### DIFF
--- a/content/en/users/tutorials/adhoc/create-your-first-virtual-machine/_index.md
+++ b/content/en/users/tutorials/adhoc/create-your-first-virtual-machine/_index.md
@@ -14,13 +14,13 @@ This tutorial describes how to create your first Virtual Machine in the EGI Fede
 
 ## Step 1: Signing up
 
-Create an EGI account with [Check-in](../../aai/check-in/signup).
+Create an EGI account with [Check-in](../../../aai/check-in/signup).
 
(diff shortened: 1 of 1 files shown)


gwarf:
> Strange that this wasn't picked up in #598

It's likely due to this setting:
https://github.com/EGI-Federation/documentation/blob/main/.github/workflows/check-links.yml#L22-L26

Would probably be good to update it, as it's seems a quite idea to ensure that no changes on organisations are not breaking links.
That said we should also be sure that all false positives are excluded (meaning as due to this we may still miss some broken links)

gwarf:
If you want/have time we could use this PR to run this check on all files, and fix other issues.
Otherwise it's for later...

enolfc (approved):
:+1:

gwarf:
@sebastian-luna-valero can you remove the link to https://wiki-igi.cnaf.infn.it/twiki/bin/view/CREAM/UserGuideEMI3 ? We still have another reference, and we will later just purge this.

enolfc:
> If you want/have time we could use this PR to run this check on all files, and fix other issues. Otherwise it's for later...

I would merge now so we have this fixed. Then open a new one with fixing other broken links

PR #621, opened 2023-11-14
Fix lint-reported errors on GitHub actions

gwarf:
Fix some lint-reported errors on GitHub actions.
By default yamllint will check keys for truthy: reporting useless errors, like for GitHub actions: https://yamllint.readthedocs.io/en/stable/rules.html#module-yam
llint.rules.truthy


enolfc (approved):
👍

PR #625, opened 2024-01-12
Update external link to operations portal documentation

gwarf:
Drop legacy link to wiki.egi.eu, and point to the wiki at IN2P3.


paolini78 (approved):
thank you

PR #642, opened 2024-03-12
Fine tune the role for cASO

enolfc:
# Summary

Be more precise on the role needed for cASO

---

**Related issue :**


CatalinCondurache (approved):
Looks good, approved

PR #649, opened 2024-04-18
Keep adjusting the quota everywhere

enolfc:
# Summary

Make sure the numbers are consistent everywhere
(Should this be just places in a single place?)

---

**Related issue :**


glarocca (approved):
LGTM

PR #667, opened 2024-08-28
Replace port number 31443

EGI-ILM:
Port 31443 does not work any more and it has been replaced by 443. The documentation has been changed accordingly

### DIFF
--- a/content/en/users/compute/orchestration/im/cli/_index.md
+++ b/content/en/users/compute/orchestration/im/cli/_index.md
@@ -43,7 +43,7 @@ parameters:
 
 ```ini
 [im_client]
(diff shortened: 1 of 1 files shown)


enolfc (approved):
👍 Thanks!

enolfc:
@gwarf is there a way to avoid the spell checker inside code blocks?

enolfc:
Forcing merge as spelling errors are from command outputs and should not be changed

gwarf:
> @gwarf is there a way to avoid the spell checker inside code blocks?

Don't know, would have to check docs/conf. Merging as you did is fine with me.

PR #671, opened 2024-09-09
Update WLCG scope tag procedure

gregcorbett:
# Summary

Remove references to REBUS in favour of reference to CRIC - as REBUS is outdated and CRIC should be used.


gwarf (approved):
LGTM

PR #682, opened 2024-11-04
Replace static table with new cvmfs-monitor

sebastian-luna-valero:
# Summary

@jose-caballero what do you think?

---

**Related issue :**


jose-caballero:
Do not replace it. The table includes links to the organisations/experiments, which are not included in the monitoring.

sebastian-luna-valero:
ok @jose-caballero but would it be ok to link to the [CVMFS monitor](http://cvmfs-release01.gridpp.rl.ac.uk/cvmfsmonitor/) after the static table?

on the other hand, it would also be a good idea to show the links to the organisations/experiments in the dynamic table shown by the [CVMFS monitor](http://cvmfs-release01.gridpp.rl.ac.uk/cvmfsmonitor/)

jose-caballero:
adding the link should be ok

jose-caballero:
Making changes for Confluence is a nightmare. All those built-in checks are more a problem than a solution.

gwarf:
> Making changes for Confluence is a nightmare. All those built-in checks are more a problem than a solution.

Confluence?

We are moving to a new check stack, and it's the one of the first real PR with it, so we have to fix some initial setup.

But we think it's helpful to have checks, even if it can need some effort indeed at some point in time :)

gwarf (approved):
LGTM.
I merged various fixes, to address issues with the new mega-linter setup.

PR #691, opened 2025-01-09
Add required permissisons to deploy previews

gwarf:
The actions should only be triggered when a label is set, and only selected users can set labels.

As per https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests, `contents: write` should be required to push the content to the preview branch, and `pull-requests: write` should be required for writing the comment.

### DIFF
--- a/.github/workflows/deploy_pr_preview.yml
+++ b/.github/workflows/deploy_pr_preview.yml
@@ -13,7 +13,9 @@ on:
     workflows: ["Build pull request preview"]
     types: [completed]
 
-permissions: read-all
+permissions:
+  contents: write
+  pull-requests: write
 
 jobs:
   deploy_pr_preview:


enolfc:
Not sure why this is still not correct:
https://github.com/EGI-Federation/documentation/actions/runs/12687215115/workflow#L16

Does it need to be merged first?

enolfc (approved):
LGTM

gwarf:
> Not sure why this is still not correct: https://github.com/EGI-Federation/documentation/actions/runs/12687215115/workflow#L16
> 
> Does it need to be merged first?

According to https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run I guess it is: `This event will only trigger a workflow run if the workflow file exists on the default branch.`

PR #685, opened 2024-12-17
Move online storage options cloud and HTC

enolfc:
# Summary

Online Storage will focus on block storage, other types of storage go to cloud or HTC
Keep redirects so there are no broken links

---

**Related issue :**

PR #664, opened 2024-07-22
Create service accounts in EGI Check-in

sebastian-luna-valero:
# Summary

Add a page on how to create service accounts, following discussion via:
https://ggus.eu/?mode=ticket_info&ticket_id=161888#update#10

---

**Related issue :** 
**Business Requirement**: [EGIREQ-156](https://jira.egi.eu/browse/EGIREQ-156)

PR #692, opened 2025-01-09
chore: adjust delete PR workflow name and set required permissions

gwarf:
I think the deletion of PR preview never works, let's see if we can get it working.

### DIFF
--- a/.github/workflows/build_pr_preview.yml
+++ b/.github/workflows/build_pr_preview.yml
@@ -8,8 +8,7 @@ on:
     # Run when label is added or present and when pushing to the PR
     types: [labeled, opened, synchronize]
 
-permissions:
-  contents: write
+permissions: read-all
 
 jobs:
   build_preview:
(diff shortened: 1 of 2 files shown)


enolfc (approved):
LGTM, but I guess we need to test again by merging

gwarf:
Hm, there may still be issues, apparently it worked when adding/removing the label but not when merging the PR:
- Success when removing a label (AFAIU): https://github.com/EGI-Federation/documentation/actions/runs/12688246198/job/35364361677
- Error when merging (AFAIU): https://github.com/EGI-Federation/documentation/actions/runs/12688986914/job/35366700297

While the workflow file is apparently the same for both run, the permissions logged in the jobs are different:
Screenshot 2025-01-09 at 14 54 32

vs 

Screenshot 2025-01-09 at 14 54 42

gwarf:
Another different is that the working was for the PR https://github.com/EGI-Federation/documentation/pull/664, made in a feature branch in this repo while the error is for the PR https://github.com/EGI-Federation/documentation/pull/685 made in a branch in your fork, which may have an impact on access to secrets.
I will log an issue and stop speaking alone in this closed PR 😅

PR #693, opened 2025-01-09
chore: remove useless parameter for EndBug/add-and-commit

gwarf:
The proper branch is checked out by the checkout action


enolfc (approved):
LGTM

PR #695, opened 2025-01-09
chore: disable spell check workflow on push, only use it in PR

gwarf:
There is no real need to spend cycles doing a spell check when pushing to main, especially if there won't be follow up on failures.

### DIFF
--- a/.github/workflows/mega-linter.yml
+++ b/.github/workflows/mega-linter.yml
@@ -4,10 +4,7 @@
 name: MegaLinter
 
 on:
-  # Trigger mega-linter at every push. Action will also be visible from Pull Requests to main
-  push:
-    branches:
-      - "main"
+  # Trigger mega-linter only on PR against main branch
   pull_request:
     branches:
       - "main"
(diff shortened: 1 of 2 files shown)


gwarf:
I think we could also disable mega-linter that runs on push, and only keep the actions that are building/deploying to GitHub pages.
Or if we keep something it should create issues/PR on errors.
But for now it's mainly consuming resources for nothing.
What do you think?

enolfc (approved):
LGTM

enolfc:
> I think we could also disable mega-linter that runs on pushes to main, and only keep the actions that are building/deploying to GitHub pages. Or if we keep something it should create issues/PR on errors. But for now it's mainly consuming resources for nothing. What do you think?

+1, should we add it here as well?
@brucellino is also reviewing the whole process so I expect more changes in this area soon

gwarf:
> > I think we could also disable mega-linter that runs on pushes to main, and only keep the actions that are building/deploying to GitHub pages. Or if we keep something it should create issues/PR on errors. But for now it's mainly consuming resources for nothing. What do you think?
> 
> +1, should we add it here as well?

I've just updated the conf of mega-linter.

>@brucellino is also reviewing the whole process so I expect more changes in this area soon

Great, I'm looking forward to this!

PR #700, opened 2025-02-14
Update GGUS section following launch of new GGUS

gregcorbett:
The GGUS documentation, at least with regards to registration, seems out of date. This PR updates the documentation with the current process (or at least my understanding of it) for accessing / registering as a support in the new GGUS system.

### DIFF
--- a/content/en/providers/operations-manuals/operations-start-guide/_index.md
+++ b/content/en/providers/operations-manuals/operations-start-guide/_index.md
@@ -309,15 +309,9 @@ requests, to be sure they are considered on time.
 
 ### Registering into GGUS
 
-To register into GGUS please follow the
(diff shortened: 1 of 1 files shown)


paolini78:
thanks for creating this PR. Indeed we have to update the whole GGUS documentation on docs.egi.eu and we have to agree this with Pavel. In the meantime, some documentation of the new system was added to this confluence space https://confluence.egi.eu/pages/viewpage.action?pageId=278856856 but we would like that all the information is reported on docs.egi.eu.
Concerning the roles in the new system, for the moment they are managed in the helpdesk itself, but for the future we would like the roles to be managed through Check-in.

paolini78 (approved):
thanks, it looks good.

PR #705, opened 2025-02-27
Tutorial updates

sebastian-luna-valero:
# Summary

Update tutorials with the latest ones from https://www.egi.eu/trainings-and-webinars/

---

**Related issue :**

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -8,6 +8,7 @@ delims
 diracsgm
 findstr
 galdyn
+geodata
 getstate
 glast
 hmac
@@ -26,6 +27,7 @@ pravda
 prettierrc
 reponame
 rsa
+Sars-Cov2
 scp
 seadatanet
 sgm
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -27,7 +27,6 @@ pravda
 prettierrc
 reponame
 rsa
-Sars-Cov2
 scp
 seadatanet
 sgm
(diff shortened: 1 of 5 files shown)


enolfc (commented):
are you completely removing old tutorials? I guess they are covered by newer versions but not sure if we want to keep some historic info

sebastian-luna-valero:
I am only removing the old tutorials that should be replaced with updated information.

Or do we want to keep everything? i.e. a summary of cloud compute in 2020, and the updated version of 2025

enolfc:
> I am only removing the old tutorials that should be replaced with updated information.
> 
> Or do we want to keep everything? i.e. a summary of cloud compute in 2020, and the updated version of 2025

Probably better to keep just the new ones so people don't get confused. Disregard my initial comment

sebastian-luna-valero:
There is an error saying that https://im.egi.eu/ does not respond, but it does... not sure how to solve it.

glarocca (approved):
LGTM

sebastian-luna-valero:
There is an error saying that https://im.egi.eu/ does not respond, but it does... not sure how to solve it.

enolfc:
> There is an error saying that https://im.egi.eu/ does not respond, but it does... not sure how to solve it.

There is also this one https://docs.google.com/presentation/d/1Qo_Zpqe9MT6X6s7o2Mz_8PplET28_WAf/edit#slide=id.p1 which is not accessible publicly

sebastian-luna-valero:
I think the only broken link now is to IM again?

enolfc:
> I think the only broken link now is to IM again?

Yes, I assume this is a false positive, so merging

PR #713, opened 2025-03-26
Update GGUS user guide

paolini78:
With the last GGUS update https://confluence.egi.eu/display/EGIHLPDSK/Update+Notes there were a few changes in the interface and I have to adjust the guide....

I've updated the text, but I still need to replace a screenshot.


damla-egi (approved):
Looks good

PR #723, opened 2025-05-20
New Keycloak VO and Group Management documentation, including visual …

EGI-ILM:
# Summary

This PR contains the new documentation for **Check-in's Group Management in Keycloak**  + some improvements on **the styles**.

## Style changes

There are some general changes that affect all pages. These have been implemented directly in `egi.css`, and include:
- More space for titles (H1, H2...) and some increase of font size, as it was difficult to distinguish them hierarchically.
- More spaces between images and text, as they were too crowded.

Additionally, other changes will become general, but will not affect existing pages:
- An option to render small images (icons)
- An option to render screenshots so that the text inside the screenshot does not get confused with the actual documentation text
- Better style for notes, so they are really highlighted. To avoid changes in other pages, a new style has been created for this purpose

These have been implemented under a minimum effort basis, so maybe there is a better way to implement them. For example, I used `important` in a couple of places, which works well buit is something I do not normally like.

A change in `config.toml` has been done so that some of these CSS classes could be easily applied in the MD.

As these changes affect the whole documentation, I have had a quick view to see that nothing is broken. It may be worth having a look at different parts of the docs as well (especially the less conventional parts).

## New Keycloak documentation
The documentation has been written from scratch, with new supportive images and new screenshots that are bigger and more readable. Small changes in the structure have been made.

**Notes:**
- The name "Keycloak" has been used, but maybe it is preferable to use another name specific to the RCIAM flavour of Keycloak.
- The section "Authorisation" could be more suitable for the "For Providers" section, rather than "For Users"
- In my opinion, the "Check-in" menu option is unnecessary. As we only have one option under "Authentication & Authorization" I would merge both in the same root menu option.
image
- Some images look too big and may need to be rescale in future but I did not want to delay this anymore
- The pre-hook formatter has messed up some of the MD and this has to be fixed for the docs to be displayed correctly.

## Next changes

- The API should be documented
- Perun docs should be improved to stay at the same level as the Keycloak documentation

PR #724, opened 2025-05-20
chore: fix #690

### LINKED ISSUE #690: The prettier pre-commit hook is deprecated -- it should be removed

We declare the prettier hook in our pre-commit configuration. However, this hook has been deprecated since April 2024 and the repository hosting it is archived.
The hook still works, but it will probably not be reliable going forward.

There does not seem to be an alternative prettier hook for the pre-commit framework; prettier themselves do not include it in their documentation.

# Summary of proposed changes

I do _not_ propose we follow the Prettier documentation on pre-commit hooks, which invoke a different framework than what we have adopted. 
Rather, I suggest we adopt a pre-commit hook configure for the content in question. 
~Unfortunately, MegaLinter is too big for a pre-commit hook.~

**Update**: Megalinter incremental has been used in other contexts and executes in a reasonable time.

Megalinter is the canonical set of policies for this repository, but the configuration is currently expressed in the Actions workflow steps as local environment variables.
These cannot be consumed by pre-commit hooks.

Therefore, the tasks at hand are:

1. Remove prettier pre-commit hook
2. Add megalinter-incremental hook with reference to repository configuration file.
3. Create repository megalinter configuration file from declarations in github actions
4. Update actions to read the configuration file rather than step environment variables

brucellino:
# Summary

This removes prettier pre-commit hook from the pre-commit configuration and adds the megalinter incremental hook with configuration taken from the github actions.

**Related issue :** #690 

Closed #690


brucellino:
I would like to get this merged asap it is blocking #723

enolfc (approved):
lgtm

brucellino:
@enolfc since you approved this can you please merge?

PR #743, opened 2025-06-19
chore: ignore markdown codeblocks in spelling

brucellino:
# Summary

We are using CSpell to spell check the content.
It is reasonable to disable spell checking for markdown codeblocks, since maintaining words in these fenced blocks will create a lot of toil.

This adds a configuration to cspell to ignore codeblocks.


enolfc (approved):
This was so much needed! 
LGTM

brucellino:
:shipit:

PR #746, opened 2025-06-20
fix(users): deal with broken links

enolfc:
# Summary

Fixes some of the broken links left by #736

---

**Related issue :**


lychee.toml
```diff
@@ -60,7 +60,7 @@ insecure = false
 # Only test links with the given schemes (e.g. https).
 # Omit to check links with any other scheme.
 # At the moment, we support http, https, file, and mailto.
-scheme = ["https"]
+scheme = ["https", "file"]
```
enolfc, line 63: @brucellino this was skipping internal links so I believe we should have this included

PR #748, opened 2025-06-20
fix: improve the looks of the signup warning box

enolfc:
# Summary

---

**Related issue :**


enolfc:
This is not achieving what I wanted, so closing

PR #762, opened 2025-08-21
Update Check-in sign up link

EGI-ILM:
As discussed internally, the link https://aai.egi.eu/signup is better for the signup process


EGI-ILM:
(Note: @FerGonzalezPerez was going to make the change, but as I need to share the docs now with someone, I have requested the change myself)

PR #774, opened 2025-09-24
Document the destination for SSM

enolfc:
# Summary

We were missing the destination queue in the SSM configuration, which has caused problems lately

---

**Related issue :**


CatalinCondurache (approved):
LGTM

PR #815, opened 2026-06-26
 providers/operations-manuals: Update man04 documentation to clarify audience.

gregcorbett:
# Summary

Clarifies audience of EGI broadcasts sent as part of man04


paolini78 (approved):
thanks for fixing this.

